PR #29, opened 2020-04-19
replace PosixPath with PurePosixPath (fixes #28)

### LINKED ISSUE #28: NotImplementedError: cannot instantiate 'CPath' on your system
I get this error when importing `my.reddit` on Windows:
```
  File "<stdin>", line 1, in <module>
  File "c:\users\migue\documents\github\hpi\my\reddit.py", line 119, in <module>
    def _get_events(backups: Sequence[Path]=get_sources(), parallel: bool=True) -> Iterator[Event]:
  File "c:\users\migue\documents\github\hpi\my\reddit.py", line 20, in get_sources
    res = list(map(CPath, files)); assert len(res) > 0
  File "C:\Users\migue\Anaconda3\lib\pathlib.py", line 1013, in __new__
    % (cls.__name__,))
NotImplementedError: cannot instantiate 'CPath' on your system
```

miguelrochefort:


karlicoss:
Hey, thanks for reporting & fix attempt! I feel likeI need to add Windows to CI. Should be straightforward in github actions..

However I tested locally (linux) and it fails to open compressed files for me. As far as I recall, the reason I used `PosixPath` was because:

- couldn't use `PurePath`/`PurePosixPath` because I needed to override `open` behaviour (which is first defined [in `Path`](https://github.com/python/cpython/blob/3.8/Lib/pathlib.py#L1210) )
- couldn't use `Path` itself because it's sealed

I'll investigate further, but at the moment seems that the quickest portable way would be to dispatch the Path flavor we need based on OS. I gave it a shot in https://github.com/karlicoss/HPI/pull/new/cpath-windows, can you let me know if it helps? 

P.S. I also need to figure out how to make [this bit](https://github.com/karlicoss/HPI/blob/b911796c151c9fd641829db6f28976e9c99d6c92/my/reddit.py#L16-L21) configurable.. not sure if you want to use `.xz` at all?

miguelrochefort:
I added [I added a comment](https://github.com/karlicoss/HPI/pull/31#issuecomment-616221985) in your other PR.

> P.S. I also need to figure out how to make this bit configurable.. not sure if you want to use .xz at all?

I'm not sure either. I haven't tried compression yet. Things seem to be working for now.

PR #31, opened 2020-04-19
Portable CPath

karlicoss:
fixes https://github.com/karlicoss/HPI/issues/28

### DIFF
--- a/my/kython/kompress.py
+++ b/my/kython/kompress.py
@@ -1,7 +1,8 @@
 """
 Various helpers for compression
 """
-from pathlib import Path, PosixPath
+import pathlib
+from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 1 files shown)


miguelrochefort:
This fixes the CPath issue, but I now get the same encoding error I had with rexport. 

Specifying the encoding fixes it:
`return kopen(str(self), encoding="utf-8") `

karlicoss:
Ah, I see! Put it [in a different place](https://github.com/karlicoss/HPI/pull/31/commits/ab61a957013ca33e0cf9e348b38a9fab7e494921) (because otherwise it breaks for archived files). Let me know if it works and I'll merge.

miguelrochefort:
It works.

PR #45, opened 2020-05-09
Better configs: safer and self documented

karlicoss:
Also some thinking on how we can make them backwards compatible too.


my/reddit.py
```diff
@@ -1,26 +1,104 @@
...
-from pathlib import Path
+
+from typing import NamedTuple, Optional
+from .core.common import PathIsh
+
+class reddit(NamedTuple):
+    '''
+    Reddit module uses [[rexport][https://github.com/karlicoss/rexport]] output
+    '''
+    export_path: PathIsh                   # path to the exported data
```
karlicoss, line 12: Use Paths; describe what it means (as well as PathIsh?)

my/google/takeout/paths.py
```diff
@@ -1,10 +1,27 @@
...
+from dataclasses import dataclass
+from ...core.common import Paths
+
+from my.config import google as user_config
+@dataclass
+class google(user_config):
+    takeout_path: Paths # path/paths/glob for the takeout zips
+###
+
+# TODO rename 'google' to 'takeout'? not sure
```
PatMyron, line 14: prefer `google` over `takeout`

Issue #67, opened 2020-07-24
Error "KeyError: 'date'" when testing my.pdfs.


thetomcraig:
I'm seeing this error when trying to test `my.pdfs` on a local PDF file.  Here's the full traceback:
```
>>> pdfs.get_annots(Path('./Ousterhout_2018_A philosophy of software design.pdf'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tomcraig/.pyenv/versions/3.7.0/envs/hpi/lib/python3.7/site-packages/my/pdfs.py", line 94, in get_annots
    return [as_annotation(raw_ann=a, path=str(p)) for a in annots]
  File "/Users/tomcraig/.pyenv/versions/3.7.0/envs/hpi/lib/python3.7/site-packages/my/pdfs.py", line 94, in <listcomp>
    return [as_annotation(raw_ann=a, path=str(p)) for a in annots]
  File "/Users/tomcraig/.pyenv/versions/3.7.0/envs/hpi/lib/python3.7/site-packages/my/pdfs.py", line 61, in as_annotation
    dates = d['date']
KeyError: 'date'
```
It appears that the input to `get_annots` expects the input, `raw_ann` to have a `date` attribute that it does not have.  I put in a print line and found that this is the value of `raw_ann`, coming from my input file;
```
{
    'page': <my.config.repos.pdfannots.pdfannots.Page object at 0x1113cd780>,
    'tagname': 'Highlight',
    'contents': None,
    'rect': [
        72, 334.555, 539.829, 420.955
    ],
    'author': 'Tom Craig',
    'text': 'In\t an\t ideal\t world,\t each\t module\t would\t be\t completely independent\tof\tthe\tothers:\ta\tdeveloper\tcould\twork\tin\tany\tof\tthe\tmodules\twithout knowing\tanything\tabout\tany\tof\tthe\tother\tmodules.\tIn\tthis\tworld,\tthe\tcomplexity of\ta\tsystem\twould\tbe\tthe\tcomplexity\tof\tits\tworst\tmodule. Unfortunately,\t this\t ideal\t is\t not\t achievable ',
    'boxes': [
        (220.556, 406.555, 539.753, 420.955),
        (72, 389.275, 539.829, 403.675),
        (72, 371.995, 539.741, 386.395),
        (72, 354.715, 400.028, 369.115),
        (93.6, 334.555, 340.65, 348.955)
    ]
}
```

[...]

PR #68, opened 2020-07-24
Add .get to call for d[date]

thetomcraig:
See issue here: https://github.com/karlicoss/HPI/issues/67.


karlicoss:
Thanks, agree it's reasonable!

PR #74, opened 2020-08-16
Add "filelist" parameter to annotated_pdfs

thetomcraig:
This PR adds a new parameter, `filelist` to the `annotated_pdfs` function.  
In my usage of this package I found that I wanted to be able to call `annotated_pdfs` with an arbitrary list of strings, containing file paths of PDF file on disk.  This allows for some arbitrary filtering upstream with unix tools to produce a list of PDF files, then the ability to pass that to `my.pdfs` and have it extract the annotations.  Currently, as far as I can tell, the only way to do this was to change the config's `roots` value, but as that is for a list of _directories_ and not a list of _files_, I added this feature.

Also added some tests for the pdfs.


karlicoss:
Thanks! 

> This allows for some arbitrary filtering upstream with unix tools to produce a list of PDF files

Yep, agree that it's a very useful scenario!

This particular module it pretty old! If I was doing this today, I'd probably use a single parameter in the config (e.g. `config.filelist`), and then use it in conjunction with [get_files](https://github.com/karlicoss/HPI/blob/master/tests/get_files.py) function. This would allow it to be pretty flexible:

- instead of `roots`, `filelist` could simply be a list of globs
- it could be any arbitrary function(`@property`), that could do pretty much anything, including calling arbitrary Unix tools
  
`is_ignored` could be removed, and the user would just directly filter `filelist`.

Either way, for now the way you changed it is ok, and later the code can be changed to use `config.filelist` in a backwards compatible fashion.

Thanks for the tests too, really appreciate :+1:

PR #76, opened 2020-09-06
fix spelling mistakes

seanbreckenridge:
Couple spelling mistake fixes patched back from my fork. These are all in comments.

Theres another spelling mistake [here](https://github.com/karlicoss/HPI/blob/07dd61ca6ae2b6de20d6954ca1584accede8b762/my/coding/commits.py#L41), (`commited_dt` -> `committed_dt`) but its part of the NamedTuple definition, so I didn't change it; just wanted to let you know incase you wanted to do something about it. (perhaps a `@property` wrapper to maintain backwards compatability? maybe thats pushing it)

### DIFF
--- a/README.org
+++ b/README.org
@@ -69,7 +69,7 @@ This library is my attempt to achieve this vision.
   - Accessing exercise data
   - Book reading progress
   - Messenger stats
-  - Querying Roam Reasearch database
+  - Querying Roam Research database
 - How does it get input data?
 - Q & A
   - Why Python?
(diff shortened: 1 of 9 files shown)


karlicoss:
Thanks!

> perhaps a @property wrapper to maintain backwards compatability?

Yep, I think this is the way to go with such things. This particular module is probably used by very few people, but I guess best not to break it unless absolutely necessary.

Also a deprecation warning would be nice, I was thinking of using  something like https://github.com/tantale/deprecated library, or just vendorizing a small, simple decorator to avoid an extra core dependency.

PR #88, opened 2020-09-30
supress conflicting regex warning

seanbreckenridge:
docs: https://docs.pytest.org/en/stable/warnings.html#warns


karlicoss:
Oh nice, thanks. I guess I was thinking suppressing user warnings across all tests altogether, so they don't mix with legitimate warnings, but assert is useful anyway!

PR #90, opened 2020-09-30
parse sms messages from xml files

seanbreckenridge:
pretty isolated change and already present here, so relatively easy to PR this.

not sure about the `datetime`s, have been thinking about that in general.

I have a [helper function](https://github.com/seanbreckenridge/HPI/blob/3d95c787153ce0d3b9e63f399f4293401fbb7439/my/core/time.py#L21-L26) to parse epoch times like this, but not sure if I should instead be putting it in local timezone/leaving it in UTC. UTC seems safe for now.

### DIFF
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -1,11 +1,10 @@
 """
 Phone calls and SMS messages
 """
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import NamedTuple, Iterator, Set, Tuple
 
-import pytz
 from lxml import etree # type: ignore
 
 from .core.common import get_files
@@ -15,6 +14,7 @@ from my.config import smscalls as config
 
 class Call(NamedTuple):
     dt: datetime
+    dt_readable: str
     duration_s: int
     who: str
 
@@ -30,6 +30,7 @@ def _extract_calls(path: Path) -> Iterator[Call]:
         # ok, so readable date is local datetime, changing throughout the backup
         yield Call(
             dt=_parse_dt_ms(cxml.get('date')),
+            dt_readable=cxml.get('readable_date'),
             duration_s=int(cxml.get('duration')),
             who=cxml.get('contact_name') # TODO number if contact is unavail??
             # TODO type? must be missing/outgoing/incoming
@@ -51,6 +52,7 @@ def calls() -> Iterator[Call]:
 
 class Message(NamedTuple):
     dt: datetime
+    dt_readable: str
     who: str
     message: str
     phone_number: str
@@ -75,14 +77,18 @@ def _extract_messages(path: Path) -> Iterator[Message]:
     for mxml in tr.findall('sms'):
         yield Message(
             dt=_parse_dt_ms(mxml.get('date')),
+            dt_readable=mxml.get('readable_date'),
             who=mxml.get('contact_name'),
             message=mxml.get('body'),
             phone_number=mxml.get('address'),
             from_me=mxml.get('type') == '2',  # 1 is received message, 2 is sent message
         )
 
+
+# See https://github.com/karlicoss/HPI/pull/90#issuecomment-702422351
+# for potentially parsing timezone from the readable_date
 def _parse_dt_ms(d: str) -> datetime:
-    return pytz.utc.localize(datetime.utcfromtimestamp(int(d) / 1000))
+    return datetime.fromtimestamp(int(d) / 1000, tz=timezone.utc)
 
 
 def stats():
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -1,15 +1,14 @@
 """
 Phone calls and SMS messages
 """
-# TODO extract SMS as well? I barely use them though..
 from datetime import datetime
 from pathlib import Path
-from typing import NamedTuple, Iterator, Set
+from typing import NamedTuple, Iterator, Set, Tuple
 
 import pytz
 from lxml import etree # type: ignore
 
-from .common import get_files
+from .core.common import get_files
 
 from my.config import smscalls as config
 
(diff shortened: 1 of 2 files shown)


seanbreckenridge:
Thought about naming it `SMS` instead of `Message` but the function is then `sms`/`smss`/`SMSs`, felt it caused confusion with the namedtuple definition.

my/smscalls.py
```diff
@@ -49,3 +47,48 @@ def calls() -> Iterator[Call]:
...
+        yield Message(
+            dt=_parse_dt_ms(mxml.get('date')),
+            who=mxml.get('contact_name'),
+            message=mxml.get('body'),
+            phone_number=mxml.get('address'),
+            from_me=mxml.get('type') == '2',  # 1 is received message, 2 is sent message
+        )
+
+def _parse_dt_ms(d: str) -> datetime:
+    return pytz.utc.localize(datetime.utcfromtimestamp(int(d) / 1000))
```
karlicoss, line 85: Wonder why I wasn't using `datetime.fromtimestamp(ts, tz=pytz.utc)` in the first place across my code.. should be equivalent, but a bit more straightforward.

seanbreckenridge: I believe `datetime.fromtimestamp` converts it to your own timezone, while `datetime.utcfromtimestamp` converts it to a datetime unaware utc timestamp.

If thats the case and you provide UTC, it should still use UTC.

At that point, it makes more sense to just do

```
>>> from datetime import datetime, timezone
>>> datetime.fromtimestamp(1601577294, tz=timezone.utc)
datetime.datetime(2020, 10, 1, 18, 34, 54, tzinfo=datetime.timezone.utc)
```

removing `pytz` altogether, and just using the builtin `datetime.timezone` objects.

Im unsure on why one would use `pytz` as well. If you don't have a reason here I can swap it to the stdlib version.

karlicoss:
Thanks, looks good!

> not sure if I should instead be putting it in local timezone/leaving it in UTC. UTC seems safe for now

Generally I'm trying to keep the original timezone if it's available from the data, UTC if it makes sense, and without the timezone if it is not available from the data. In this case I'm actually a bit confused -- judging by my data export, it seems that the timestamps never change (regardless whether the export was in winter or summer), so I assume it's always UTC? However, the human readable dates displayed on my phone are also the same as in XML files -- somehow feels weird if SMS/calls datetimes change when you change timezones .. unless Android isn't keeping track of the timezone information.

This is a somewhat unfortunate thing about living in London timezone -- harder to catch such issues :) You're in the US, right, does UTC look correct for your data?

my/smscalls.py
karlicoss: Huh, interesting, somehow wasn't aware of `timezone.utc`! Don't mind replacing, thanks.

seanbreckenridge:
Yeah, Im in the US. I'll take a look at the export later today and see how it compares.

seanbreckenridge:
Just as an update.

So, looks like the date has two attributes related to that, heres an extract of the date related info:

```
<sms protocol="0"  date="1599927139924" type="1"  date_sent="1599927138000" sub_id="1" readable_date="Sep 12, 2020 9:12:19 AM" />
```

`date` is in UTC, so thats still a possibility to return. But, I believe it would be possible to parse the readable date, and see what the difference is between two datetime unaware objects to get the timezone. proof of concept:

```python
>>> from datetime import datetime, timezone
>>> epoch = 1599927139
>>> readable_date = "Sep 12, 2020 9:12:19 AM"
>>> utc_time = datetime.utcfromtimestamp(epoch)  # tz unaware, but in UTC
>>> local_time = datetime.strptime(readable_date, "%b %d, %Y %H:%M:%S %p")  # tz unaware, is 'wrong' by timezone offset
>>> int((local_time - utc_time).total_seconds() / 3600)
-7
```

readable date doesn't give anything related to daylight savings. I'm not sure what would happen if the phone changed timezones, or how this handles daylight savings.

Its possible to just have a map from `time zone offset` -> `some timezone` in that, but that may not always be correct.

I think it should always expose at least the epoch time (int), and perhaps store the readable date as well? Could have a `property` wrapper that attempts to parse it like above, but not even sure how useful that is.

You can always sort by the epoch time instead of trying to use some timezone aware time (would just cause more issues if you're trying to sort by timezone aware time). And if you want it to be readable, `readable_date` gives the info in whatever timezone you were in.

Maybe could also/instead include a `timezone` `@property` wrapper that does what the code above does instead? That just returns an hour offset, so you can do whatever you want with UTC time + parsed offset.

Thoughts?

karlicoss:
> Its possible to just have a map from time zone offset -> some timezone in that, but that may not always be correct.

Yep! Actually it's a somewhat common pattern, some data sources only use local time, and some only return UTC and have no tz info.

And I think I have a solution: if you have location history, you basically have your 'timezone history'. So if you have a tz-unaware/UTC date, you can ask the timezone provider to set it for you. 
If there is no automatic location history, a good approximation would be just a list of timezones in the config, along with the dates when you moved in between them throughout you life (hopefully only a few datapoints).
And I have this thing implemented already, it's just in the unpublished bit of my HPI at the moment, but I was planning to merge it soonish. 

A reverse lookup for timezones might be useful anyway, but yeah, considering it's ambiguous might cause downstream issues with DST and similar things.

Otherwise, IMO defaulting on UTC would make the most sense for now, as for readable datetime, up to you!

Joshfindit:
If I may suggest: store not only the original value, but the location of the device when it received messages.
With those, it would make it possible to retroactively correct timestamp interpretation.

seanbreckenridge:
I don't see location of the device as part of the export (`my.smscalls` uses [sms backup & restore](https://play.google.com/store/apps/details?id=com.riteshsahu.SMSBackupRestore&hl=en_US)). That would probably be a separate phone application, and HPI module, and then perhaps merged into this data somehow.

seanbreckenridge:
Changes:
 - Ended up using the builtin `datetime.timezone.utc`
 - attached `dt_readable` as a string to each `Call`/`Message`, incase you want to 'pretty print' or have access to the string to print.
 - Linked to this issue at the parse function, incase someone wants to try and extract a timezone

Issue #17, opened 2020-04-12
Support different error handling policies


karlicoss:
As described [here](https://beepb00p.xyz/mypy-error-handling.html#global_policy). 

E.g. might make sense to implement:

- fail fast, i.e. throw instead if yielding exceptions
- fully defensive, i.e. ignore exceptions instead of yielding

That would require cooperation from either underlying data providers, or we could decorate them or something. Perhaps, makes more sense to do it on the HPI level.

karlicoss:
Timezone resolution is perhaps one place where it would make a lot of sense -- in many case it's not super bad to fallback on UTC or use local timezone. But tricky to make it generic enough.

karlicoss:
an example in `google_takeout_parser`: https://github.com/seanbreckenridge/google_takeout_parser/blob/bd6a40390de1784bf48bac4a71bb4558f9d61cdc/google_takeout_parser/path_dispatch.py#L153

PR #106, opened 2020-10-31
my.time.tz: implement different policies for localizing

karlicoss:
related: https://github.com/karlicoss/HPI/issues/17

also follow up for https://github.com/karlicoss/HPI/pull/96#issuecomment-713833765

### DIFF
--- a/my/time/tz/common.py
+++ b/my/time/tz/common.py
@@ -1,0 +1,41 @@
+from datetime import datetime
+from typing import Callable, cast
+
+from ...core.common import tzdatetime, Literal
+
+
+'''
+Depending on the specific data provider and your level of paranoia you might expect different behaviour.. E.g.:
+- if your objects already have tz info, you might not need to call localize() at all
+- it's safer when either all of your objects are tz aware or all are tz unware, not a mixture
+- you might trust your original timezone, or it might just be UTC, and you want to use something more reasonable
+'''
+Policy = Literal[
+    'keep'   , # if datetime is tz aware, just preserve it
+    'convert', # if datetime is tz aware, convert to provider's tz
+    'throw'  , # if datetime is tz aware, throw exception
+    # todo 'warn'? not sure if very useful
+]
+
+def default_policy() -> Policy:
+    try:
+        from my.config import time as user_config
+        return cast(Policy, user_config.tz.policy)
+    except Exception as e:
+        # todo meh.. need to think how to do this more carefully
+        # rationale: do not mess with user's data until they want
+        return 'keep'
+
+
+def localize_with_policy(lfun: Callable[[datetime], tzdatetime], dt: datetime, policy: Policy=default_policy()) -> tzdatetime:
+    tz = dt.tzinfo
+    if tz is None:
+        return lfun(dt)
+
+    if   policy == 'keep':
+        return dt
+    elif policy == 'convert':
+        ldt = lfun(dt.replace(tzinfo=None))
+        return dt.astimezone(ldt.tzinfo)
+    else: # policy == 'error':
+        raise RuntimeError(f"{dt} already has timezone information (use 'policy' argument to adjust this behaviour)")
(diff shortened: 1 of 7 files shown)


my/time/tz/common.py
```diff
@@ -0,0 +1,34 @@
...
+- you might trust your original tiemzone, or it might just be UTC, and you want to use something more reasonable
+'''
+Policy = Literal[
+    'keep'   , # if datetime is tz aware, just preserve it
+    'convert', # if datetime is tz aware, convert to provider's tz
+    'throw'  , # if datetime is tz aware, throw exception
+    # todo 'warn'? not sure if very useful
+]
+# todo make the default configurable, e.g. add my.cofnig.time.tz section
+DEFAULT_POLICY: Policy = 'keep'
```
karlicoss, line 20: not sure what the default should be, 'keep' or 'convert'? @seanbreckenridge what do you think? 
I'm not sure because I'm expecting to use this in 'error' mode most of the time to fail fast :) 
An upside of `convert` would be is that you basically put trust into the `tz` module, which is ultimately good. A downside is that it might be too coarse (e.g. if `tz` fell back on the `home` locations), in that case the original timezone might be better.

my/time/tz/common.py
```diff
@@ -0,0 +1,34 @@
...
+from ...core.common import tzdatetime
+
+
+'''
+Depending on the specific data provider and your level of paranoia you might expect different behaviour.. E.g.:
+- if your objects already have tz info, you might not need to call localize() at all
+- it's safer when either all of your objects are tz aware or all are tz unware, not a mixture
+- you might trust your original tiemzone, or it might just be UTC, and you want to use something more reasonable
+'''
+Policy = Literal[
```
karlicoss, line 13: Not sure about `Literal`, haven't really used it much before.. But in my experience having real enums for such things just makes it a bit annoying (e.g. you have to import it first, etc). I guess mypy gives enough safety while keeping it flexible.
Either way should be easy to switch in the future, so maybe good to give it a go anyway.

my/time/tz/common.py
seanbreckenridge: I assume this will will be configurable through `my.config`. I think it makes sense to have the default be `keep`, forcing the user to be aware of the `tz` module and configure it makes changes to the data they may not be expecting.

seanbreckenridge:
wonderful :+1:

```python
In [1]: from more_itertools import last

In [2]: from my.time.tz.main import localize

In [3]: from my.body import weight

In [4]: localize(last(weight()).when)
Out[4]: datetime.datetime(2020, 10, 31, 15, 18, 57, tzinfo=<DstTzInfo 'America/Los_Angeles' PDT-1 day, 17:00:00 DST>)
```

Not sure if you want to add this as a utility function to HPI directly, but I'll probably add a hacky function like:

```python
X = TypeVar("X")
def localize_iterable(it: Iterator[X]) -> Iterator[X]:
	"""
	If policy == convert, localize all datetimes on some iterable sequence
	"""
	if policy != 'convert':
		yield from it
		return
	for item in it:
		if hasattr(it, 'dt'):
			# tz.localize and yield
		elif hasattr(it, 'at'):
			# tz.localize and yield
		elif hasattr(it, 'when'):
			# tz.localize and yield
		else:
			warnings.warn(f"Couldnt find datetime attr {it}")
			yield tz
```

Seeing as I have multiple attributes that match datetimes, and I prefer 'at'/'when' to 'dt' sometimes.

May be easier to check the definition of 'X' to see if the constructor has a typed attribute with the type `datetime`, unsure.

karlicoss:
Oh, nice! Good idea, I'll think about it too.

> May be easier to standarize 'dt' across all modules (or use @property wrappers), or check the definition of 'X' to see if the constructor has a typed attribute with the type datetime, unsure.

Yeah, or a special decorator, or type alias (which acts like `datetime`, but indicates that it's an 'event time'). Or all of these :)
In principle, it would be very nice to apply them uniformly without having to modify the data source code, would be simpler and better decomposition.

Also this feels like it's begging for lenses :) Haven't really used them in other languages before either, but found something for Python https://github.com/ingolemo/python-lenses

PR #132, opened 2021-03-14
cleanup coding.commits

seanbreckenridge:
remove the _things check, its never activated
for me and seems pointless

update mechanism for finding fdfind/fd path,
send a core.warning if it fails

update mechanism to cache repos to new cachew
api (remove hashf), cache repos on a per-repo
basis

### DIFF
--- a/my/coding/commits.py
+++ b/my/coding/commits.py
@@ -108,9 +108,7 @@ def canonical_name(repo: Path) -> str:
 
 
 def _fd_path() -> str:
-    fd_path: Optional[str] = (
-        shutil.which("fdfind") or shutil.which("fd-find") or shutil.which("fd")
-    )
+    fd_path: Optional[str] = shutil.which("fdfind") or shutil.which("fd-find") or shutil.which("fd")
     if fd_path is None:
         high(f"my.coding.commits requires 'fd' to be installed, See https://github.com/sharkdp/fd#installation")
     # TODO: this just causes it to fail if 'fd' can't be found, but the warning is still sent... seems fine?
--- a/my/coding/commits.py
+++ b/my/coding/commits.py
@@ -2,11 +2,17 @@
 Git commits data for repositories on your filesystem
 """
 
+import shutil
+import string
 from pathlib import Path
 from datetime import datetime, timezone
 from typing import List, NamedTuple, Optional, Dict, Any, Iterator, Set
 
-from ..common import PathIsh, LazyLogger, mcachew
+from ..core.common import PathIsh, LazyLogger, mcachew, Stats
+from ..core.cachew import cache_dir
+from ..core.warnings import high
+
+# TODO: create user_config dataclass?
 from my.config import commits as config
 
 # pip3 install gitpython
@@ -17,23 +23,12 @@ from git.repo.fun import is_git_dir, find_worktree_git_dir # type: ignore
 log = LazyLogger('my.commits', level='info')
 
 
-_things = {
-    *config.emails,
-    *config.names,
-}
-
-
 def by_me(c) -> bool:
     actor = c.author
     if actor.email in config.emails:
         return True
     if actor.name in config.names:
         return True
-    aa = f"{actor.email} {actor.name}"
-    for thing in _things:
-        if thing in aa:
-            # TODO this is probably useless
-            raise RuntimeError("WARNING!!!", actor, c, c.repo)
     return False
 
 
@@ -43,8 +38,8 @@ class Commit(NamedTuple):
     message: str
     repo: str # TODO put canonical name here straightaway??
     sha: str
-    ref: Optional[str]=None
-        # TODO filter so they are authored by me
+    ref: Optional[str] = None
+    # TODO filter so they are authored by me
 
     @property
     def dt(self) -> datetime:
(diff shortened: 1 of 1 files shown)


my/coding/commits.py
```diff
@@ -112,18 +107,27 @@ def canonical_name(repo: Path) -> str:
     #     pass # TODO 
 
 
+def _fd_path() -> str:
```
karlicoss, line 110: guess later could move this to core, could be useful for other modules too

my/coding/commits.py
```diff
@@ -112,18 +107,27 @@ def canonical_name(repo: Path) -> str:
...
+    fd_path: Optional[str] = shutil.which("fdfind") or shutil.which("fd-find") or shutil.which("fd")
+    if fd_path is None:
+        high(f"my.coding.commits requires 'fd' to be installed, See https://github.com/sharkdp/fd#installation")
+    # TODO: this just causes it to fail if 'fd' can't be found, but the warning is still sent... seems fine?
```
karlicoss, line 114: I guess here it should be assert since it won't work otherwise anyway? (it would also make mypy happy without `or "fd"`)
In principle, there could be a fallback on `walk` or something later, similar to promnesia https://github.com/karlicoss/promnesia/blob/e44a2bda053cc93bb8dc1b84fd1a4461aa5d914f/src/promnesia/common.py#L437

my/coding/commits.py
```diff
@@ -2,11 +2,17 @@
...
 from pathlib import Path
 from datetime import datetime, timezone
 from typing import List, NamedTuple, Optional, Dict, Any, Iterator, Set
 
-from ..common import PathIsh, LazyLogger, mcachew
+from ..core.common import PathIsh, LazyLogger, mcachew, Stats
+from ..core.cachew import cache_dir
+from ..core.warnings import high
+
+# TODO: create user_config dataclass?
```
karlicoss, line 15: Yeah.. I guess as for schema,
- either of emails or names could be optional (but it should warn/throw if both are set to None?)
- for `roots` would be nice to allow to specify a single path, kind of like `Paths/get_files` handling does.. but here the discovery of 'files' is somewhat different, so maybe not worth the trouble and it's fine as a `Sequence[PathIsh]` for now

karlicoss (approved):
Nice, thanks for the cleanup! Looks good to me, just a minor comment, but happy to merge as is, and fix that later.

my/coding/commits.py
```diff
@@ -139,35 +143,43 @@ def repos():
...
-        yield from repo_commits(r)
+        yield from _cached_commits(r)
+
+_allowed_letters: str = string.ascii_letters + string.digits
+
+
+def _cached_commits_path(p: Path) -> str:
+    # compute a reduced simple filepath using the absolute path of the repo
+    simple_path = ''.join(filter(lambda c: c in _allowed_letters, str(p.absolute())))
+    return cache_dir() / simple_path / '_cached_commits'
```
karlicoss, line 171: Maybe worth having a subdirectory as well (i.e. `cache_dir() / 'commits'), otherwise it pollutes the top level cache dir. Not a problem, but makes it a bit more confusing

seanbreckenridge: Yeah, I could reverse it so that `_cached_commits` is the top-level subdirectory and each repo gets a folder in there.

Maybe I should note that this 'simplified path' is something [I used to have in core](https://github.com/seanbreckenridge/HPI/blob/6e2262d66b332a83099e3740aa3559204427e95b/my/core/file.py#L76-L84), because it was also useful for `my.google`, to quickly be able to create a [`cache_path`](https://github.com/seanbreckenridge/HPI/blob/6e2262d66b332a83099e3740aa3559204427e95b/my/coding/commits.py#L185-L192) based on a simplified version of the filepath. I don't know if you've run into this issue before but I often did when parsing large GDPR exports. Maybe its worth to have a helper for this?

karlicoss: > Yeah, I could reverse it so that _cached_commits is the top-level subdirectory and each repo gets a folder in there.
I'm about to submit a PR with some more cleanup, was gonna share with you!

> Maybe its worth to have a helper for this?
Yep, sure!

my/coding/commits.py
seanbreckenridge: Yeah, I suppose an assert is a better error message than the long 'command not found'

Not sure how much overlap you have with helper methods like these in HPI and promnesia, not sure if it'd be too much work to split some parts of core into a separate package like `exporthelpers` or if that would just cause more complications.

my/coding/commits.py
seanbreckenridge: Sound like sensible defaults, sure

I've typically just used email, as I feel theres a non-zero chance that someone else has the same name as me in some giant project that may be cloned in my repos

my/coding/commits.py
karlicoss: Yeah, not sure either -- with exporthelpers figured it was worth it since it was shared among so many (10+?) projects. And still, used submodules instead since maintaining & releasing extra package on pypi is kinda pain.
I guess two copies of code I can kind of tolerate for now -- maybe if it matures and later if HPI core is extracted it would make sense to just use it in Promnesia regardless whether any HPI modules are used.

seanbreckenridge: Yeah, thats about what I expected, makes sense

PR #133, opened 2021-03-14
smscalls: add config block

seanbreckenridge:
so that don't have to infer what
to set in your hpi config based
on usage in module

### DIFF
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -1,15 +1,26 @@
 """
 Phone calls and SMS messages
+Exported using https://play.google.com/store/apps/details?id=com.riteshsahu.SMSBackupRestore&hl=en_US
 """
+
+from .core import PathIsh, dataclass
+from my.config import smscalls as user_config
+
+@dataclass
+class smscalls(user_config):
+    # directory that SMSBackupRestore syncs XML files to
+    export_path: PathIsh
+
+from .core.cfg import make_config
+config = make_config(smscalls)
+
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import NamedTuple, Iterator, Set, Tuple
 
 from lxml import etree # type: ignore
 
-from .core.common import get_files
-
-from my.config import smscalls as config
+from .core.common import get_files, Stats
 
 
 class Call(NamedTuple):
@@ -91,7 +102,7 @@ def _parse_dt_ms(d: str) -> datetime:
     return datetime.fromtimestamp(int(d) / 1000, tz=timezone.utc)
 
 
-def stats():
+def stats() -> Stats:
     from .core import stat
 
     return {


my/smscalls.py
```diff
@@ -1,15 +1,26 @@
...
+from .core import PathIsh, dataclass
+from my.config import smscalls as user_config
+
+@dataclass
+class smscalls(user_config):
+    # directory that SMSBackupRestore syncs XML files to
+    export_path: PathIsh
+
+from .core.cfg import make_config
+config = make_config(smscalls)
```
karlicoss, line 15: In principle, `make_config` is only necessary either if there is a config 'migration', or default arguments, otherwise it's possible to simply use the class attributes. But I guess doesn't hurt it for consistentcy

karlicoss (approved):

karlicoss:
Thanks!

my/smscalls.py
seanbreckenridge: ah, ok; I tend to just add them to the config blocks everywhere

I suppose if you don't have a migration you're supporting its not necessary, but other than the bit of extra code and tiny performance hit I don't think theres much of a down side

Also, perhaps in this case, `smscalls` may change somehow, because it isn't immediately obvious that you provide just the base directory, and the `calls-*.xml` and `sms-*.xml` are the defaults that are assumed

karlicoss: > it isn't immediately obvious that you provide just the base directory

Ah indeed -- I try to avoid hardcoding it in general (since `get_files` supports globs), but here wasn't sure -- since there is a mix of files. Maybe it could just rely on a single `get_xmls` function, and then calls/messages can just filter filenames explicitly based on `calls`/`sms` substrings? Since that's basically what's given by SMSBackupRestore tool, and unlikely to change

Also I guess a 'more correct' type for `export_path` here is `Paths`, since it's anything that's supported by `get_files`

seanbreckenridge: > Since that's basically what's given by SMSBackupRestore tool, and unlikely to change

I think I agree, that its probably a sensible default...

I think ideally one would be able to provide either:

* just the base directory, and the globs are assumed (what we're currently doing)
* globs for sms and calls individually, which are then processed by get_files, like:

```python
class smscalls:
	export_path: Optional[Paths]
    call_xml_paths: Optional[Paths]
    sms_xml_paths: Optional[Paths]
```

I'm not a fan of the attribute names (`call_xml_paths`) or that everything is `Optional`, but that would give the user the most freedom.

---

For now, perhaps just put a TODO [`here`](https://github.com/karlicoss/HPI/blob/master/my/smscalls.py#L9) mentioning the possible issue, and if someone wants to open an issue for it/if they have a different way of organizing the XML files, it can be tackled when its an issue?

karlicoss: Ah -- you think that someone might split them into separate directories/rename? Yeah I guess it's possible. 

Another option for is making the module friendlier to user override, so exposing `get_calls_files` and `get_messages_files`. Kind of like `def inputs` in some other modules.

seanbreckenridge: Until someone mentions it being a problem, I think its fine to leave as is.

PR #135, opened 2021-03-14
add similar projects to README

almereyda:
In preparation of an attempt to regain autonomy of my very own personal data, I have been researching into the field in the past years and consider these two additions important enough to propose them here.

- Perkeep, formerly known as Camlistore, is a content-addressable storage for personal data that has been around for long and also seeks a wider adoption base. People interested in the HPI might also find a very useful infrastructure in that.
- Open Humans is an online community from dedicated researchers that aim at easing the process of obtaining and analysing certain available personal data sets. Their infrastructure includes surveys as well as interactive notebooks with secure access to a personal data store and API integrations for receiving archives from third party services.

### DIFF
--- a/README.org
+++ b/README.org
@@ -689,6 +689,8 @@ Similar projects:
 - [[https://github.com/bcongdon/bolero][bcongdon/bolero]]: exposes your personal data as a REST API
 - [[https://en.wikipedia.org/wiki/Solid_(web_decentralization_project)#Design][Solid project]]: personal data pod, which websites pull data from
 - [[https://remotestorage.io][remoteStorage]]: open protocol for apps to write data to your own storage
+- [[https://perkeep.org]][Perkeep]: a tool with [[https://perkeep.org/doc/principles]][principles] and esp. [[https://perkeep.org/doc/uses]][use cases] for self-sovereign storage of personal data
+- [[https://www.openhumans.org]][Open Humans]: a community and infrastructure to analyse and share personal data
 
 Other links:
 


karlicoss:
Thanks!
I tried perkeep a couple of times, but it didn't click, guess will give it another shot!
And yeah, for Open Humans I have plans to plug in some notebooks!

seanbreckenridge:
Have had a similar issue with perkeep -- I think it was what I was trying to get work before I found HPI.

It feels like it handles a lot of the lower level and complex problems with storage management, but storage space hasn't really been an issue, and syncthing works well enough most of the time for me

At that point I think the project was somewhat unmaintained https://twitter.com/PerkeepOrg/status/1095371716644155392

But it does seem they've had releases recently

I think its a great piece of software, Im just not sure how I'd use it

karlicoss:
Yeah, I managed to import some data in PerKeep but then I wasn't sure how to access it and what can I do with it. I'll watch some talks/demos perhaps

almereyda:
Ah, thanks for the pointer. Especially due to its lower-level simplicity and awareness for possibilities of sovereignity in distributed storage environments is why I like Perkeep. Of course they would need a wider roll out to validate their hypotheses at large. Some integration with NAS distributions comes to mind, or well-described Aarch64 home lab storage cluster setups.

We could also mention Tahoe-LAFS in the same vein, and the dev. was in Berlin for dtn #2 dtn.is and made sure they still also do care about exactly those subjects, of self-storage, distribution schematics, in a social sense: of what is actually possible and achievable in the world outside of developer circles, and the dawn of the fading archives and pasts, for not will history be written by the winners (here: large institutions able to invest into perseverance and preservation) again.

Issue #46, opened 2020-05-10
Discussion: configuring modules


karlicoss:

seanbreckenridge:
Just some thoughts on configuring/modifying this.

This is heavily biased as someone who is familiar with python, and I expected that I'd have to fork/modify HPI to fit my needs.

In general, I'm happy with configuration/tooling that this has provided around a layer to load data. Splitting up the configuration into `~/.config/my` and using dataclasses to validate attributes makes sense once I've seen it a few times.

Though, when configuration is programmatic (python), and theres a lack of documentation/dataclass (e.g. [`my.coding.commits`](https://github.com/karlicoss/HPI/blob/master/my/coding/commits.py)), its unclear what exactly to do. I was able to figure out what fields were necessary by inference, but in a larger namespace/module that might not be obvious.

(sidenote; I believe its this, feel free to correct me if I'm wrong)

```
class commits:
    names: List[str] = ["Sean Breckenridge"]
    emails: List[str] = ["seanbrecke@gmail.com", "email two...."]
    roots: Paths = [Path(environ["REPOS"])];
```

I'd think that everyone has different modules/needs. Not everyone is going to use the same hardware, social media, or systems for note taking, so if your objective is to have programmatic access to all of your data, I don't think `pip install hpi` would work in most cases. Either `pip install -e .` or `with_my` seem like fine solutions to someone who wants to heavily configure/modify this. 

I feel like using HPI is closer to the process of curating a vim/emacs/dotfiles configuration. This provides a useful starting point and the exporters (`rexport`/`ghexport`) can be plugged in for commonly used modules. Then (optionally), you write your own custom modules for your own data sources. Its slightly more annoying in this case, because of all the parsing/normalization required for ever-changing APIs/Data-Sources, but when/if exporters/DALs are somewhat standardized/pushed to pypi, I think a configuration file/object in `~/.my/config` that specifies which modules are active/inactive - I'm a fan of how [doom emacs](https://github.com/seanbreckenridge/dotfiles/blob/master/.config/doom/init.el#L19) does this - could be useful. Similar to doom/spacemacs, each module/namespace package could (optionally) have a `doctor.py` which would do checks, `model.py` for the `NamedTuple` representation, and/or `__init__.py`/`<modulename>.py`, which would be the python module one would import from the REPL to get access to the data. 

[...]

PR #136, opened 2021-03-15
my.coding.commits: test/mypy on CI + cleanup

karlicoss:


seanbreckenridge:
Great, looks good :+1:

my/coding/commits.py
```diff
@@ -1,38 +1,64 @@
...
+
+from my.core import PathIsh, LazyLogger, make_config
+from my.core.cachew import cache_dir
+from my.core.common import mcachew
+from my.core.warnings import high
+
+
+from my.config import commits as user_config
+@dataclass
+class commits_cfg(user_config):
```
karlicoss, line 26: name conflict with `def commits` below! But I guess doesn't hurt since it's basically internal to this module

my/coding/commits.py
```diff
@@ -1,38 +1,64 @@
...
+    roots: Sequence[PathIsh] = field(default_factory=list)
```
karlicoss, line 27: Interesting that without it, it would complain
`ValueError: mutable default <class 'list'> for field roots is not allowed: use default_factory` -- so would require you to use `tuple` in the user config. Didn't know about that.. but good that relatively easy to workaround with this field trick.. Not sure if there is an easier way (maybe some `@dataclass` annotation attributes).

my/coding/commits.py
```diff
@@ -1,38 +1,64 @@
...
-from ..core.common import PathIsh, LazyLogger, mcachew, Stats
-from ..core.cachew import cache_dir
-from ..core.warnings import high
 
-# TODO: create user_config dataclass?
-from my.config import commits as config
+# experiment to make it lazy?
+# would be nice to have a nicer syntax for it... maybe make_config could return a 'lazy' object
+@functools.lru_cache()
+def config() -> commits_cfg:
```
karlicoss, line 35: Not sure about lru_cache in particular, but thinking that 'lazy' instantiation like this might be good to use in general, often the fact that config is constructed on the top level makes it harder to override/test.

Guess relevant to https://github.com/karlicoss/HPI/issues/46

my/coding/commits.py
```diff
@@ -168,7 +195,7 @@ def _commits(_repos: List[Path]) -> Iterator[Commit]:
 def _cached_commits_path(p: Path) -> str:
     # compute a reduced simple filepath using the absolute path of the repo
     simple_path = ''.join(filter(lambda c: c in _allowed_letters, str(p.absolute())))
-    return cache_dir() / simple_path / '_cached_commits'
+    return str(cache_dir() / 'commits' / simple_path / '_cached_commits')
```
karlicoss, line 198: wonder why mypy hasn't caught missing `str` before!

my/coding/commits.py
seanbreckenridge: Hmm..

I'd think one might run into less problems if you avoid  `lru_cache` for things like config, because it means the user has less options to do things like monkey patching.

It does look/feel worse to have to handle caching the result of some computation after the config has been loaded. I think previously some of that had been done in `@property` functions on the config itself, but that doesn't feel like the best solution.

I think I agree with your comment:

> would be nice to have a nicer syntax for it... maybe make_config could return a 'lazy' object

seanbreckenridge: Another way that I've often handled this is to have an [`inputs` function](https://github.com/seanbreckenridge/HPI/blob/master/my/ttt.py#L39-L52) which takes what the config gives you, and computes the resulting inputs for each of the functions. Then that function is the [default parameter for any function](https://github.com/seanbreckenridge/HPI/blob/master/my/ttt.py#L70) which requires the input files

Is a bit confusing on first glance, but is nice for testing purposes and being able to specify a subset of the paths if I want to do something in the repl

karlicoss: Hmm yeah ok, good point about caching!
I guess it doesn't hurt to instantiate the config multiple times, negligible in comparison with the rest of computation

my/coding/commits.py
seanbreckenridge: Ah, strange. I think initially I didn't have this split into a separate function, it was just a lambda in the `@cachew` decorator, and there its fine as `cachew` expects a [`PathProvider`](https://github.com/karlicoss/cachew/blob/master/src/cachew/__init__.py#L715)? but `mypy` should've definitely caught that this wasn't returning a Path...

my/coding/commits.py
karlicoss: > the default parameter for any function which requires the input files

Yeah, I thought about that too, but then it often requires propagating the whole thing down the call stack.. I guess it's a classic 'state' problem

Issue #138, opened 2021-03-15
Add core.query/serialize?


seanbreckenridge:
Not sure if this is something you'd be interesting in having in the core here/in utils, is the leftover helper modules I created while maintaining my fork.

As I was writing [`HPI_API`](https://github.com/seanbreckenridge/HPI_API) I added a `core.serialize` and `core.query` to my HPI fork as well.

It is quite magical (as it just resolves the function name [with a string](https://github.com/seanbreckenridge/HPI/blob/master/my/utils/query.py#L16-L19)) but it lets me do some simple queries pretty easily, and play around with pipelines in the shell without having to worry about how to interop with python/dumping something from the REPL

https://github.com/seanbreckenridge/HPI/blob/master/my/utils/query.py
https://github.com/seanbreckenridge/HPI/blob/master/my/utils/serialize.py

and then a script that exposes that info:

https://github.com/seanbreckenridge/HPI/blob/master/scripts/hpi_query

As some examples, 5 songs I listened to recently:

```
$ hpi_query my.mpv history | jq -r '.[] | .path' | grep -i 'music' | head -n 5
/home/sean/Music/Radiohead/1994 - The Bends/11 - Sulk.mp3
/home/sean/Music/Radiohead/1994 - The Bends/02 - The Bends.mp3
/home/sean/Music/Nujabes/Nujabes - Metaphorical Music (2003) [V0]/10. Next View.mp3
/home/sean/Music/Earth, Wind & Fire/Earth Wind And Fire - Greatest Hits - [MP3-V0]/16 - After The Love Has Gone.mp3
/home/sean/Music/Darren Korb/Darren Korb - Transistor Original Soundtrack[2013] (V0)/14 - Darren Korb - Apex Beat.mp3
```

I also use this in my menu bar, to print how many calories/water I've drank today: 

![image](https://user-images.githubusercontent.com/7804791/111110050-15b82b80-8519-11eb-9574-4cf370f1f261.png)

Like:

```
#!/bin/bash
# how many water I've had today

((BLOCK_BUTTON == 3)) && notify-send "$("${REPOS}/HPI/scripts/water-recent")"

HPI_QUERY="${REPOS}/HPI/scripts/hpi_query"
{
	"${HPI_QUERY}" --days 1 'my.food' 'water' | jq -r '.[] | .glasses'
	echo 0
} | datamash sum 1
```

Have even had some fun creating graphs like this in the terminal:

```
[...]

PR #140, opened 2021-03-18
my.core.serialize: orjson with additional default and _serialize hook

seanbreckenridge:
Not sure if this is ready to be merged yet, just creating the PR so that this tests can run for `my.core.serialize`, and for possible feedback on the second commit here.

Perhaps `my.core.query` will be a separate PR, as this has already caused a bunch of changes

Not sure how you'd like to list `orjson` in the dependencies? I added it to `optional`, as thats where `logzero` is as well. But, needed to also add it to tox, like the `pip install cachew` there.

Also, should mention I moved `error_to_json` and `to_jsons` to `my.core.pandas`, since they're now pandas-aware. I didn't maintain a copy/renamed import for backwards compatibility, but since [these are only a month](https://github.com/karlicoss/HPI/commit/4012f9b7c2a429170df8600591ec8d1e1407b162) old, I think thats fine? Else I can move them back.

Related: https://github.com/karlicoss/HPI/issues/138

### DIFF
--- a/my/core/pandas.py
+++ b/my/core/pandas.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from pprint import pformat
 from typing import Optional, TYPE_CHECKING, Any, Iterable, Type, List, Dict
 from . import warnings, Res
-from .common import LazyLogger
+from .common import LazyLogger, Json, asdict
 
 logger = LazyLogger(__name__)
 
@@ -97,8 +97,23 @@ def check_dataframe(f: FuncT, error_col_policy: ErrorColPolicy='add_if_missing',
 # todo doctor: could have a suggesion to wrap dataframes with it?? discover by return type?
 
 
-from .error import error_to_json
-error_to_row = error_to_json # todo deprecate?
+def error_to_row(e: Exception, *, dt_col: str='dt', tz=None) -> Json:
+    from .error import error_to_json, extract_error_datetime
+    edt = extract_error_datetime(e)
+    if edt is not None and edt.tzinfo is None and tz is not None:
+        edt = edt.replace(tzinfo=tz)
+    err_dict: Json = error_to_json(e)
+    err_dict[dt_col] = edt
+    return err_dict
+
+
+# todo not sure about naming
+def to_jsons(it: Iterable[Res[Any]]) -> Iterable[Json]:
+    for r in it:
+        if isinstance(r, Exception):
+            yield error_to_row(r)
+        else:
+            yield asdict(r)
 
 
 # mm. https://github.com/python/mypy/issues/8564
@@ -111,6 +126,7 @@ def _as_columns(s: Schema) -> Dict[str, Type]:
     if D.is_dataclass(s):
         return {f.name: f.type for f in D.fields(s)}
     # else must be NamedTuple??
+    # todo assert my.core.common.is_namedtuple?
     return getattr(s, '_field_types')
 
 
@@ -124,7 +140,6 @@ def as_dataframe(it: Iterable[Res[Any]], schema: Optional[Schema]=None) -> DataF
     #    https://github.com/pandas-dev/pandas/blob/fc9fdba6592bdb5d0d1147ce4d65639acd897565/pandas/core/frame.py#L562
     # same for NamedTuple -- seems that it takes whatever schema the first NT has
     # so we need to convert each individually... sigh
-    from .common import to_jsons
     import pandas as pd
     columns = None if schema is None else list(_as_columns(schema).keys())
     return pd.DataFrame(to_jsons(it), columns=columns)
--- a/my/core/serialize.py
+++ b/my/core/serialize.py
@@ -1,0 +1,40 @@
+from typing import Any
+
+from .common import is_namedtuple
+from .error import error_to_json
+
+# note: it would be nice to combine the 'asdict' and _orjson_default to some function
+# that takes a complex python object and returns JSON-compatible fields, while still
+# being a dictionary.
+# a workaround is to encode with dumps below and then json.loads it immediately
+
+
+def _orjson_default(obj: Any) -> Any:
+    """
+    Encodes complex python datatypes to simpler representations,
+    before they're serialized to JSON string
+    """
+    # orjson doesn't serialize namedtuples to avoid serializing
+    # them as tuples (arrays), since they're technically a subclass
+    if is_namedtuple(obj):
+        return obj._asdict()
+    if isinstance(obj, Exception):
+        err = error_to_json(obj)
+        # remove unrelated dt key? maybe error_to_json should be refactored?
+        err.pop('dt', None)
+        return err
+    raise TypeError(f"Could not serialize object of type {obj.__type__.__name__}")
+
+
+def dumps(obj: Any) -> str:
+    try:
+        import orjson
+        # serialize 'b'"1970-01-01T00:00:00"' instead of b'"1970-01-01T00:00:00.000000"
+        opts = orjson.OPT_OMIT_MICROSECONDS
+        json_bytes = orjson.dumps(obj, default=_orjson_default, option=opts)
+        return json_bytes.decode('utf-8')
+    except ModuleNotFoundError:
+        import warnings
+        warnings.warn("You might want to install 'orjson' to support serialization for lots more types!")
+        import json
+        return json.dumps(obj, default=_orjson_default)
--- a/my/core/serialize.py
+++ b/my/core/serialize.py
@@ -1,15 +1,20 @@
-from typing import Any
+import datetime
+from typing import Any, Optional, Callable
+from functools import lru_cache
 
 from .common import is_namedtuple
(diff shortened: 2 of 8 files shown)


seanbreckenridge:
Hmm, [I thought I added the `orjson` dep to tox](https://github.com/karlicoss/HPI/pull/140/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R26), not sure why [its failing](https://github.com/karlicoss/HPI/pull/140/checks?check_run_id=2138671251#step:5:218)

seanbreckenridge:
Ah, core runs before the dependency is installed.

I think the solution here would be to create a separate `tests/serialize.py` file and call that in `testenv:tests-all`?

seanbreckenridge:
As a sort of real world example...

Ran a quick test using my [hpi_query](https://github.com/seanbreckenridge/HPI/blob/master/scripts/hpi_query) script:

```
#!/bin/bash

query() {
  echo "$1" "$2"
  $REPOS/HPI/scripts/hpi_query "$1" "$2" --count 1 | jq
}

query my.browsing history
query my.battery history
query my.discord messages
query my.food food
query my.food food
query my.ipython history
query my.money transactions
query my.mpv history
query my.reddit comments
query my.old_forums history
query my.ttt history
query my.window_watcher history
query my.zsh history
query my.games.chess history
query my.games.league history
query my.games.steam achievements
query my.location.all exact_locations
query my.smscalls messages
query my.smscalls calls
query my.github.all events
query my.github.ghexport events
query my.github.gdpr events
```

and all of those serialized properly

tox.ini
```diff
@@ -23,6 +23,7 @@ setenv = MY_CONFIG = nonexistent
 commands =
     pip install -e .[testing]
     pip install cachew
+    pip install orjson
```
karlicoss, line 26: > I think the basic solution here would be to create a separate tests/serialize.py file and call that in testenv:tests-all?

Yeah, I guess it's good to start with... Basically in principle this belongs to `tests-core`, for `tests-all` I had in mind that it runs the 'module tests'. But also want to have a way of running tests with/without optional libraries (like `pandas`/`cachew` and `orjson`) to make sure the fallbacks work properly.
Maybe something like a parameterized test which would temporary 'shadow' optional modules from `sys.modules` could work... Or a separate tier of tests which runs core + optional libraries. Not sure, but okay to keep it this way for now I guess.

my/core/pandas.py
```diff
@@ -97,8 +97,23 @@ def check_dataframe(f: FuncT, error_col_policy: ErrorColPolicy='add_if_missing',
 # todo doctor: could have a suggesion to wrap dataframes with it?? discover by return type?
 
 
-from .error import error_to_json
-error_to_row = error_to_json # todo deprecate?
+def error_to_row(e: Exception, *, dt_col: str='dt', tz=None) -> Json:
```
karlicoss, line 100: > I didn't mainta in a copy/renamed import for backwards compatibility, but since these are only a month old, I think thats fine?

Yep, no problem!

my/core/serialize.py
```diff
@@ -0,0 +1,170 @@
...
+def _dumps_factory(**kwargs) -> Callable[[Any], str]:
+    use_default: DefaultEncoder = _default_encode
+    # if the user passed an additional 'default' parameter,
+    # try using that to serialize before before _default_encode
+    _additional_default: Optional[DefaultEncoder] = kwargs.get("default")
+    if _additional_default is not None and callable(_additional_default):
+
+        def wrapped_default(obj: Any) -> Any:
+            try:
+                # hmm... shouldn't mypy know that _additional_default is not None here?
```
karlicoss, line 51: Perhaps because the `def` doesn't capture the actual value, but only the name of the variable. So by the time it's called someone could have reassigned it back to `None`?
Maybe possible to workaround with `Final`, but not sure mypy would be smart enough to do this analysis -- probably not worth the hassle

my/core/serialize.py
```diff
@@ -0,0 +1,170 @@
...
+        # most keys are typically attributes from a NT/Dataclass,
+        # so most seem to work: https://github.com/ijl/orjson#opt_non_str_keys
+        def _orjson_dumps(obj: Any) -> str:
+            # orjson returns json as bytes, encode to string
+            return orjson.dumps(obj, **kwargs).decode('utf-8')
+
+        return _orjson_dumps
+    except ModuleNotFoundError:
+        import json
+        import warnings
```
karlicoss, line 76: could use `core.warnings` ... although I still haven't come up to some consistent patterns with using it, need to think about it later

my/core/serialize.py
```diff
@@ -0,0 +1,170 @@
...
+                return _default_encode(obj)
+
+        use_default = wrapped_default
+
+    kwargs["default"] = use_default
+
+    try:
+        import orjson
+
+        # todo: add orjson.OPT_NON_STR_KEYS? would require some bitwise ops
```
karlicoss, line 66: hmm.. yeah a bit magical, on the other hand definitely doesn't hurt in some cases. What do you mean by bitwise ops though?

my/core/serialize.py
```diff
@@ -0,0 +1,170 @@
...
+    # serialized as a dataclass or namedtuple
+    # discussion: https://github.com/karlicoss/HPI/issues/138#issuecomment-801704929
+    if hasattr(obj, '_serialize') and callable(obj._serialize):
+        return obj._serialize()
+    raise TypeError(f"Could not serialize object of type {type(obj).__name__}")
+
+
+# could possibly run multiple times/raise warning if you provide different 'default'
+# functions or change the kwargs? The alternative is to maintain all of this at the module
+# level, which is just as annoying. Maybe should be increased to lru_cache(16) or something...?
```
karlicoss, line 40: Yeah, I guess just making it unbounded doesn't hurt?

karlicoss (approved):

karlicoss:
Looks good to me, thanks! Yeah query could definitely go as a separate PR. 
Happy to merge, and I'll play with influxdb over the weekend a bit more to make sure it covers these usecases well.
Are you happy if I squash it all? (for less 'fix' commits) Or you want to do interactive rebase first perhaps before merging?

tox.ini
seanbreckenridge: > But also want to have a way of running tests with/without optional libraries (like pandas/cachew and orjson) to make sure the fallbacks work properly

Ah, thats true, I don't think I'm currently testing for that. Would require to have a test somewhere in `tests-core` (perhaps adding it directly to the  that runs before `orjson` is installed, and then different tests for when orjson is installed. Perhaps could add another test which tests the fallback and add a regular (non-starred) import in [here](https://github.com/karlicoss/HPI/blob/master/tests/core.py)

Like `from my.core.serialize import test_without_orjson_installed`?

my/core/serialize.py
seanbreckenridge: Yeah, I was considering that, I just saw that stdlib `warnings` module was still being used in the `logzero` message, so I stuck with that for now. Is safer than expecting `termcolor`/`colorama` to be installed, but if you think thats robust enough I'm fine using `my.core.warnings`.

tox.ini
karlicoss: Yeah, makes sense! and it could also be called by the orjson version as a normal function

my/core/serialize.py
seanbreckenridge: `orjson` options is an integer which gets `or`d together:

```
>>> import orjson
>>> orjson.OPT_STRICT_INTEGER
64
>>> orjson.OPT_NAIVE_UTC
2
```

like `orjson.dumps(obj, option=orjson.OPT_STRICT_INTEGER | orjson.OPT_NAIVE_UTC)`

My bitwise arithmetic is a bit rusty, but I guess thinking about it now you'd just `or` whatever option you receieved with `orjson.OPT_NON_STR_KEYS` and if it was present already that would have no effect

my/core/serialize.py
karlicoss: Yeah, it's termcolor defensive as well (yet another thing to test for optional dependencies as we discussed above). But I think fine to keep it either way -- can just go through all warnings later and clean them up

my/core/serialize.py
seanbreckenridge: unbounded would also be fine, yeah. feel free to change that

seanbreckenridge:
Feel free to squash the commits and merge

my/core/serialize.py
karlicoss: Yep, hopefully it's just a matter of bitwise or.
Thinking it's another candidate for 'defensive' policy by the way, but I think it makes sense to allow non-str keys for default yeah.

seanbreckenridge:
Added a fallback test which uses `json.dumps` during `tests-core`, and updated the `lru_cache` maxsize to `None`

Believe thats everything

karlicoss:
awesome, thanks!

PR #143, opened 2021-03-23
initial my.core.query implementation

seanbreckenridge:
This is already quite complicated, so I'll probably add a subcommand to `hpi` like my [hpi_query](https://github.com/seanbreckenridge/HPI/blob/master/scripts/hpi_query) script in another PR (perhaps while migrating it to [click](https://click.palletsprojects.com/)?; or maybe thats another ordeal on its own...)

Opening this PR so that this may be reviewed, and for possible feedback/changes

Notes:

- `parse_timedelta_string` isn't currently used, it'll eventually be used in the CLI interface to be able to parse durations like [this](https://github.com/seanbreckenridge/HPI/blob/master/scripts/hpi_query#L13-L18), so you could do something: `hpi query module function --within 6h` to query for items from that function within the last 6 hours.
- I'm a bit unsure on some of the typing, generics isn't my strong-suit, so some things may be wrong.
- Tests still need to be added for this, but I'm unsure how to do that? Since this requires some 'data source' to use query. Should I just create a 'dummy' namedtuple/dataclass in `tests/query.py` and use that? or do you already have something here I could use
- Should docs be created for this? I wrote a substantial docstring for the `select` function, so hopefully thats understandable. With some examples as tests perhaps thats good enough?
- Totally open to changes, this was a rough implementation with the first ideas that come to mind
- Though the CLI interface doesn't current exist, this works when calling from in python, e.g.:

```python
import my.browsing
import my.mpv
import my.zsh
import my.ttt
import my.reddit

from itertools import chain

# wrap items which dont have datetime as a NT/DT attribute

def wrapped_reddit():
    for r in my.reddit.comments():
        yield {
            attr: getattr(r, attr)
            for attr in
            ("created", "url", "text")
        }


# any mixed events could be chained here, if were using `order_value`
events = chain(
    wrapped_reddit(),
    my.zsh.history(),
    my.ttt.history(),
    my.mpv.history()
    my.browsing.history()

)


from my.core.query import select

timeline = select(events, order_value=lambda o: isinstance(o, datetime))

count = 0
for item in timeline:
    count += 1
    print(item)

print(count)
```

As some more basic examples

```python
In [8]: list(select(("my.zsh", "history",), limit=5))
Out[8]:
[Entry(dt=datetime.datetime(2020, 5, 18, 8, 23, 22, tzinfo=datetime.timezone.utc), duration=0, command='z'),
 Entry(dt=datetime.datetime(2020, 5, 18, 8, 23, 22, tzinfo=datetime.timezone.utc), duration=0, command='en env_config.zsh'),
 Entry(dt=datetime.datetime(2020, 5, 18, 8, 23, 22, tzinfo=datetime.timezone.utc), duration=0, command='ls'),
 Entry(dt=datetime.datetime(2020, 5, 18, 8, 23, 22, tzinfo=datetime.timezone.utc), duration=0, command='ec'),
 Entry(dt=datetime.datetime(2020, 5, 18, 8, 23, 12, tzinfo=datetime.timezone.utc), duration=0, command='ls')]

In [9]: next(select(("my.zsh", "history",), where=lambda c: "ffmpeg" in c))
Out[9]: Entry(dt=datetime.datetime(2020, 9, 7, 7, 31, 49, tzinfo=datetime.timezone.utc), duration=0, command='ffmpeg')
```

```python
In [1]: import my.zsh; from datetime import date; td = date.today()

In [3]: # more analysis like, how many zsh commands today

In [4]: # can also pass function without calling

In [7]: ilen(select(my.zsh.history, where=lambda o: o.dt.date() == td))
Out[7]: 2107

In [16]: from my.body import food

In [17]: # thing with most calories?

In [18]: next(select(food, order_key="calories", limit=1, reverse=True))
Out[18]: Food(when=datetime.datetime(2020, 11, 30, 2, 47, 7, tzinfo=datetime.timezone.utc), food='barbeque corn chips', calories=1125)
```


seanbreckenridge:
Related issue: https://github.com/karlicoss/HPI/issues/138

karlicoss:
> parse_timedelta_string isn't currently used, it'll eventually be used in the CLI interface to be able to parse durations like this, so you could do something: hpi query module function --within 6h to query for items from that function within the last 6 hours.

Oh nice -- definitely useful. I'm usually using [`pandas.to_timedelta`](https://pandas.pydata.org/docs/reference/api/pandas.to_timedelta.html) in ad-hoc applications. Guess we want to keep it decoupled from pandas, but guess good to keep it compatible with it? (it is now, just for the future). But perhaps this won't need to be more complicated than it's now.

> Tests still need to be added for this, but I'm unsure how to do that? Since this requires some 'data source' to use query. Should I just create a 'dummy' namedtuple/dataclass in tests/query.py and use that? or do you already have something here I could use

Yeah -- I've added some test data repos as submodules: https://github.com/karlicoss/HPI/blob/master/.gitmodules#L1-L6

For example, these run on CI, against 'real' data
https://github.com/karlicoss/HPI/blob/1cdef6f40a6e4c0dcb545a1f9e0e592f909301f3/tests/bluemaestro.py#L41-L46
https://github.com/karlicoss/HPI/blob/1cdef6f40a6e4c0dcb545a1f9e0e592f909301f3/tests/location.py#L35-L38

But perhaps for this would make sense to test on less specific things (since it doesn't need modules even, just takes iterables) Yeah, like dummy dataclasses.


> Should docs be created for this? I wrote a substantial docstring for the select function, so hopefully thats understandable. With some examples as tests perhaps thats good enough?

Yeah, that's awesome thanks. Hmm maybe it's about time to start `doc/FEATURES.org` or something like that? Could at least mention it and point at the docstring/tests, probably no need for too much duplication. Maybe later it would make sense to write docstrings for the most useful core functions and generate some automatic html docs for it, or somathing like that.

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+
+    if order_by is not None or order_key is not None or order_value is not None:
+        # we have some sort of input that specifies we should reorder the iterator
+
+        order_by_chosen: Optional[OrderFunc] = order_by  # if the user just supplied a function themselves
+        if order_by is None:
+            # https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.spy
+            [first_item], itrc = more_itertools.spy(itr)
+            # replace the 'itr' in the higher scope with itrc -- itr is consumed by more_itertools.spy
+            itr = itrc
```
karlicoss, line 326: I usually just write `[first_item], itr = spy(itr)`. But guess it might be cryptic to someone who sees spy for the first time :)

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+    if where is not None:
+        itr = filter(where, itr)
+
+    if order_by is not None or order_key is not None or order_value is not None:
+        # we have some sort of input that specifies we should reorder the iterator
+
+        order_by_chosen: Optional[OrderFunc] = order_by  # if the user just supplied a function themselves
+        if order_by is None:
+            # https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.spy
+            [first_item], itrc = more_itertools.spy(itr)
```
karlicoss, line 324: might crash if the iterable is empty?

seanbreckenridge:
> But perhaps this won't need to be more complicated than it's now.

Yeah, I dont see any usecases to query `ms`/`us` or anything smaller. Perhaps `week` and `month` should be added, though I haven't found myself needing to use them often. Most of the times I find myself using `--days 1` or `--days 7`, so the user can just do `--within 30d` to do a month; isn't that difficult. Also means we dont have to handle [leap years](https://github.com/seanbreckenridge/glue/blob/master/lib/glue/date_utils.ex#L18-L19) or dealing with dates more than necessary

> But perhaps for this would make sense to test on less specific things

Yeah, I agree -- this just needs to have some function/iterable to query against, so best to remove the additional point of failure

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+            itr = more_itertools.always_reversible(itr)
+
+    # apply limit argument
+    if limit is not None:
+        return itertools.islice(itr, limit)
+
+    return itr
+
+
+
```
karlicoss, line 374: ideas for test:
- something involving mix of `Exception` and normal items to check that `Res` entries are handled properly
- something involving rogue entries for which predicates can't be computed -- guess what should happen in this case would depending on drop/raise parameters?

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+            # try to use a key, if it was supplied
+            # order_key doesn't use local state - it just tries to find the passed
+            # attribute, or default to the 'default' value. As mentioned above,
+            # best used for items with a similar structure
+            if order_key is not None:
+                order_by_chosen = _generate_order_by_func(first_item, key=order_key, default=default)
+                if order_by_chosen is None:
+                    raise QueryException(f"Error while ordering: could not find {order_key} on {first_item}")
+            elif order_value is not None:
+                itr1, itr2 = itertools.tee(itr, 2)  # expensive!!!
```
karlicoss, line 336: I guess in principle this doesn't matter that much since `sorted` later would need linear memory anyway.
And sorting also will have at least linear slowdown, so the linear time spent computing `order_by_lookup` also isn't a big problem

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+    this object returns the value to order by
+    """
+    if isinstance(obj_res, Exception):
+        if default is not None:
+            return lambda _o: default
+        else:
+            low(f"""While creating order_by function, encountered exception {obj_res}
+Value to order_by unknown, provide a 'default', filter exceptons with a 'where' predicate or
+pass 'drop_errors' to ignore this""")
+            return lambda _o: None
```
karlicoss, line 113: This would probably raise when actually trying to compare? Perhaps could use some sort of special 'terminal object' which always compares `False` or `True` (depending on whether we want these items to show in the beginning or in the end). This might also make `default` argument from the `select` interface unnecessary? Although don't mind having it either -- guess just don't know how useful it would be in advance

Imo, even better, Exceptions and objects for which the predicate fails to compute in general would stick to the closest 'good' entry -- that way might be easier to debug. 
Got an experimental helper for that but it's not very generic, and either way can improve this later
 https://github.com/karlicoss/HPI/blob/eb26cf863345f764a8eb279aff0589d67e397bae/my/core/error.py#L53-L57

my/core/query.py
```diff
@@ -0,0 +1,386 @@
...
+    # and on some objects it'll return None, the getattr(o, field_name, default) won't
+    # use the default, since it finds the attribute (it just happens to be set to None)
+    # should this do something like: 'lambda o: getattr(o, k, default) or default'
+    # that would fix the case, but is additional work. Perhaps the user should instead
+    # write a 'where' function, to check for that 'isinstance' on an Optional field,
+    # and not include those objects in the src iterable
+
+    # user must provide either a key or a where predicate
+    if where_function is not None:
+        if isinstance(obj, dict):
```
karlicoss, line 145: seems like a common pattern to dispatch like this (dict/NT/dataclass) -- maybe later can extract somewhere

karlicoss (commented):
Looks awesome, just a few minor comments! 
I think types look alright -- and considering how dynamic this is, it's inevitable to have some ignore comments

my/core/query.py
seanbreckenridge: Ah thats true...

Could use `more_itertools.peekable` instead and catch the `StopIteration`?

Would mean the above comment `[first_item], itr = spy(itr)` would be resolved as well

```
>>> import more_itertools
>>> more_itertools.peekable([])
<more_itertools.more.peekable object at 0x7f9bb1afe3d0>
>>> x = more_itertools.peekable([1,2])
>>> x.peek()
1
>>> next(x)
1
>>> x.peek()
2
>>> x = more_itertools.peekable([])
>>> x.peek()
  File "<stdin>", line 1, in <module>
  File "/home/sean/.local/lib/python3.9/site-packages/more_itertools/more.py", line 311, in peek
    self._cache.append(next(self._it))
StopIteration

```

If the iterator is empty, I suppose it could return immediately? I don't think it should be an error, as its totally possible your `where` predicate filtered the `src` down to no items, which should be case thats supported.

my/core/query.py
seanbreckenridge: Agreed, yeah. Could probably be moved near `asdict` in `my.core.common`

my/core/query.py
karlicoss: yeah, probably easier to just return immediately? Agree shouldn't be error, maybe warnining so it doesn't just quietly produce no output?

my/core/query.py
seanbreckenridge: Yeah, It would probably fail with a 'cant compare NoneType' error? 

I was originally trying some way to push exceptions and None to the front of the list, as they're sort of 'unsortable', but I couldn't figure out what the 'terminal object' should be, since for example:

If the user passes an `order_value` function and it trying to find datetimes `order_value=lambda o: isinstance(o, datetime)`, this would then be trying to compare `None` to a `datetime`, but theres no way for this to know that without an additional `O(n)` check (maybe thats fine though)

At one point I had another flag to `drop_unsorted`, which removed error/NoneType before, but I thought it might be a bit too expensive and its essentially another linear computation. Perhaps that doesn't really matter though -- the ability to sort by mixed types means that its worth it.

I think perhaps its possible to add another linear check [here](https://github.com/seanbreckenridge/HPI-to-master/blob/query/my/core/query.py#L359) that calls the `order_by_chosen` functions on each item in the iterable. If any of those return None it could:

- if `drop_unsorted` is present, drop them
- else, put unsorted items at the end of the list

karlicoss: by a terminal object I meant something like this:

```
[ins] In [20]: @total_ordering 
          ...: class Dummy: 
          ...:     def __le__(self, other): 
          ...:         return True 
          ...:                                                                                                                                                                                                                                                

[ins] In [21]: dummy = Dummy()                                                                                                                                                                                                                                

[ins] In [22]: list(sorted([5,  None, 3, 9, None], key=lambda i: dummy if i is None else i))                                                                                                                                                                  
Out[22]: [None, None, 3, 5, 9]
```

So `dummy` now will always make the error entries stick to one side of the list.
I could swear something like this was in stdlib, but can't find it 🤔

seanbreckenridge: Ah, that makes sense -- would just wrap unsortable items into that.

I'll experiment with that. Should probably be enabled by default (just like how `Iterator[Res[Model]]` functions can return errors, this should also just forward errors back), with a flag to disable it (which could cause errors, but thats a feature you're enabling then...

will probably add both `wrap_unsorted` (True by default) and `drop_unsorted` (False by default)

seanbreckenridge:
Added all the changes we discussed, along with the tests (though, due to the volume/dummy data iterators, unsure if they should be here or in tests/query.py)

my/core/query.py
seanbreckenridge: Ah, I just realized. How items are determined as `Unsortable` is different if something is `order_key` or `order_val`. If you had a rogue item in `order_key`, it would just be classified as `Unsortable`, but in `order_val`, since it goes through and pre-generates order by funcs for each item, a rogue item would [cause an error here](https://github.com/karlicoss/HPI/pull/143/files#diff-13571aa0237710934b67bf68764a486c074521f32e7fe7fc04059e40a674d83bR398). Perhaps an error shouldn't be raised there, since `_handle_unsorted` below will split it into the unsorted iterable anyways?

The current behavior just raises the error:

https://github.com/karlicoss/HPI/pull/143/files#diff-13571aa0237710934b67bf68764a486c074521f32e7fe7fc04059e40a674d83bR543-R548

seanbreckenridge: Potential implementation:

```diff
diff --git a/my/core/query.py b/my/core/query.py
index 048b0ed..0583b86 100644
--- a/my/core/query.py
+++ b/my/core/query.py
@@ -70,7 +70,8 @@ def _generate_order_by_func(
         obj_res: Res[T],
         key: Optional[str] = None,
         where_function: Optional[Where] = None,
-        default: Optional[U] = None
+        default: Optional[U] = None,
+        force_unsorted: bool = False,
 ) -> Optional[OrderFunc]:
     """
     Accepts an object Res[T] (Instance of some class or Exception)
@@ -156,8 +157,12 @@ pass 'drop_exceptions' to ignore exceptions""")
         # warn here? it seems like you typically wouldn't want to just set the order by to
         # the same value everywhere, but maybe you did this on purpose?
         return lambda _o: default
-
-    return None  # couldn't compute a OrderFunc for this class/instance
+    elif force_unsorted:
+        # generate a dummy function which returns None
+        # this causes this type of object to be classified as an unsortable item
+        return lambda _o: None
+    else:
+        return None  # couldn't compute a OrderFunc for this class/instance
 
 
 def _drop_exceptions(itr: Iterator[ET]) -> Iterator[T]:
@@ -393,9 +398,9 @@ Your 'src' may have been empty of the 'where' clause filtered the iterable to no
                 for obj_res in itr1:
                     key: Any = _determine_order_by_value_key(obj_res)
                     if key not in order_by_lookup:
-                        keyfunc: Optional[OrderFunc] = _generate_order_by_func(obj_res, where_function=order_value, default=default)
-                        if keyfunc is None:
-                            raise QueryException(f"Error while ordering: could not determine how to order {obj_res}")
+                        keyfunc: Optional[OrderFunc] = _generate_order_by_func(obj_res, where_function=order_value, default=default, force_unsorted=True)
+                        # should never be none, as we have force_unsorted=True
+                        assert keyfunc is not None
                         order_by_lookup[key] = keyfunc
 
                 # set the 'itr' (iterator in higher scope)
@@ -542,10 +547,10 @@ def test_order_key_multi_type() -> None:
 
 def test_couldnt_determine_order() -> None:
 
-    import pytest
-
-    with pytest.raises(QueryException, match=r"could not determine how to order"):
-        res = list(select(iter([object()]), order_value=lambda o: isinstance(o, datetime)))
+    res = list(select(iter([object()]), order_value=lambda o: isinstance(o, datetime)))
+    assert len(res) == 1
+    assert isinstance(res[0], Unsortable)
+    assert type(res[0].obj) == object
 
 
 # same value type, different keys, with clashing keys
 ```

Comment (seanbreckenridge) [reply my/core/query.py:452]: Added a commit with a slightly modified version of above (just changed the names/added more documentation)

File: my/core/query.py [hunk 7]
```diff
@@ -0,0 +1,717 @@
...
+
+def test_parse_timedelta_string():
+
+    import pytest
+
+    with pytest.raises(ValueError) as v:
+        parse_timedelta_string("5xxx")
+
+    assert v is not None
+    assert str(v.value).startswith("Could not parse time duration from")
```
karlicoss, line 470: nit: could use `match=` argument for `pytest.raises` as well?

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+
+    import pytest
+    from collections import Counter
+
+    # by default should wrap unsortable (error)
+    with pytest.warns(UserWarning, match=r"encountered exception") as w:
+        res = list(select(_mixed_iter_errors(), order_value=lambda o: isinstance(o, datetime)))
+    assert Counter(map(lambda t: type(t).__name__, res)) == Counter({"_A": 4, "_B": 2, "Unsortable": 1})
+    # compare the returned error wrapped in the Unsortable
+    returned_error = next(filter(lambda o: isinstance(o, Unsortable), res)).obj  # type: ignore[union-attr]
```
karlicoss, line 670: not super important, but I think the ignore is needed here because mypy can't peek into the lambda. I usually use something like `(o for o in res if isinstance(r, Exception))` -- same length and mypy can guess the type correctly!

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+        for v in range(1, 6):
+            yield _Int(v)
+
+    def filter_two(obj: Any) -> bool:
+        return obj.x != 2
+
+    res = list(select(basic_iter(), where=filter_two, reverse=True))
+    assert len(res) == 4
+
+    for (actual, expected) in zip(res, (5, 4, 3, 1)):
```
karlicoss, line 503: nit, but I usually try asserting on lists instead of iterating and asserting on individual items -- gives better error messages

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+    y: int
+    z: int
+
+
+class _B(NamedTuple):
+    y: datetime
+
+
+# move these to tests/? They are re-used so much in the tests below,
+# not sure where the best place for these is
```
karlicoss, line 572: yeah -- I guess it's nice to have some basic test[s] near the code, as a sort of demonstration, but maybe the details should go in the actual test file? Although still looks OK to me, maybe with more core tests it will become clearer what's the best way

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+    yield _A(x=datetime(year=2005, month=5, day=10, hour=4, minute=10, second=1), y=10, z=2)
+    yield _A(x=datetime(year=2009, month=3, day=10, hour=4, minute=10, second=1), y=12, z=1)
+    yield _B(y=datetime(year=1995, month=5, day=10, hour=4, minute=10, second=1))
+    yield _A(x=datetime(year=2005, month=4, day=10, hour=4, minute=10, second=1), y=2, z=-5)
+
+
+def _mixed_iter_errors() -> Iterator[Res[Union[_A, _B]]]:
+    m = _mixed_iter()
+    for _ in range(3):
+        yield next(m)
```
karlicoss, line 585: nit: doesn't really matter here, but I guess in principle best to use `yield from itertools.islice(m, 0, 3)` here -- naked `next` often make error in code very confusing because they can throw `StopIteration`

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+        if isinstance(o, Unsortable):
+            continue
+        ordval = orderfunc(o)
+        if ordval is None:
+            continue
+        yield o
+
+
+# hmm... is there really not a stdlib/more_itertools function to split an iterable by a predicate?
+# similar to 'my.core.error.sort_res_by'?
```
karlicoss, line 211: yep! `https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.partition`
Although I often do something like
```
it1, it2 = tee(it)
res1 = (i for i in it1 if     isinstance(i, Unsortable))
res2 = (i for i in it2 if not isinstance(i, Unsortable))
return (res1, res2)
```
this makes it mypy-friendly, and slightly more iterative. Although in theory this still can end up taking linear memory depending on the downstream code, but so would storing them in a list, or partition -- don't think there is any way around it unless you're assuming you can iterate input iterable multiple times.

my/core/query.py
karlicoss: Oh also -- just curious for my understanding, not super important implementation wise I guess. In principle this could be lazy right? In the sense that the `order_by_lookup` cache could be computed for the item the first time it's used by the code below?

my/core/query.py
```diff
@@ -0,0 +1,717 @@
...
+import itertools
+from datetime import datetime, date, timedelta
+from typing import TypeVar, Tuple, Optional, Union, Callable, Iterable, Iterator, Dict, Any, NamedTuple, List
+
+import more_itertools
+
+from .warnings import low
+from .common import is_namedtuple
+from .error import Res, unwrap
+from .warnings import low
```
karlicoss, line 21: nit: double import?

karlicoss (commented):
looks great, and thanks for tests!
basically happy to merge, just a few minor comments, let me know if you wanna fix any of them now, otherwise can do later

my/core/query.py
seanbreckenridge: I had tried to wrap my head around how to do that - the limiting factor made me write it this way was that `sorted` needed to be passed a function, and I couldn't figure out a nice way to make it lazy, while having to pass it an flags/other info on how to sort items...

It just became too much a mess of dynamically creating additional named-functions (i.e. `def some_function`) to handle state

In principle it could be, I believe, I wasn't able to figure out a nice way to do so

my/core/query.py
seanbreckenridge: Ah, was searching for `split` on the more_itertools docs... thanks.

Its become a bit more complicated since, as it has to check for recursive `Unsortable` instances, so it wouldnt be as simple as the above anymore.

Also seeing as whenever this is called it also has to call `sorted` (which will definitely consume the iterator), there isn't much benefit to iterators

think I will just leave as it is

my/core/query.py
seanbreckenridge: yeah, makes sense

seanbreckenridge:
Github is doing something weird with me not being able to mark of one of the comments resolved since I somehow replied to the parent? Anyways, regarding making the sort function for `order_by_chosen` lazy:

I had tried to wrap my head around how to do that - the limiting factor made me write it this way was that sorted needed to be passed a function, and I couldn't figure out a nice way to make it lazy, while having to pass it an flags/other info on how to sort items...

It just became too much a mess of dynamically creating additional named-functions (i.e. def some_function) to handle state

In principle it could be, I believe, I wasn't able to figure out a nice way to do so

---

added the other changes described -- feel free to squash and merge (by the way, In general you don't have to ask to squash, for any future PRs)

karlicoss:
Awesome, thanks! :tada:

PR #152, opened 2021-04-02
core: discovery_pure; allow multiple package roots

seanbreckenridge:
iterates over `my.__path__._path` if possible
to discover additional namespace package paths
to discover modules under

else defaults to the path relative to
the current file

makes things like this possible:

```
 hpi module install --user my.discord

Running: /usr/bin/python3 -m pip install --user git+https://github.com/seanbreckenridge/discord_data
Collecting git+https://github.com/seanbreckenridge/discord_data
  Cloning https://github.com/seanbreckenridge/discord_data to /tmp/pip-req-build-a1uwx9p4
  Running command git clone -q https://github.com/seanbreckenridge/discord_data /tmp/pip-req-build-a1uwx9p4
```

specified by the block in [my HPI repo](https://github.com/seanbreckenridge/HPI/blob/master/my/discord.py)

when beforehand it failed with:

```
 hpi module install --user my.discord

/home/sean/Repos/HPI-to-master/my
  File "/home/sean/.local/bin/hpi", line 33, in <module>
    sys.exit(load_entry_point('HPI', 'console_scripts', 'hpi')())
  File "/home/sean/Repos/HPI-to-master/my/core/__main__.py", line 424, in main
    func(args)
  File "/home/sean/Repos/HPI-to-master/my/core/__main__.py", line 357, in module_install
    *_requires(module),
  File "/home/sean/Repos/HPI-to-master/my/core/__main__.py", line 331, in _requires
    mod = module_by_name(module)
  File "/home/sean/Repos/HPI-to-master/my/core/discovery_pure.py", line 155, in module_by_name
    raise RuntimeError(f'No such module: {name}')
RuntimeError: No such module: my.discord
```

### DIFF
--- a/my/core/discovery_pure.py
+++ b/my/core/discovery_pure.py
@@ -114,6 +114,17 @@ def _extract_requirements(a: ast.Module) -> Requires:
 
 # todo should probably be more defensive..
 def all_modules() -> Iterable[HPIModule]:
+    """
+    Return all importable modules under all items in the 'my' namespace package
+
+    Note: This returns all modules under all roots - if you have
+    several overlays (multiple items in my.__path__ and you've overridden
+    modules), this can return multiple HPIModule objects with the same
+    name. It should respect import order, as we're traversing
+    in my.__path__ order, so module_by_name should still work
+    and return the correctly resolved module, but all_modules
+    can have duplicates
+    """
     for my_root in _iter_my_roots():
         yield from _modules_under_root(my_root)
 
@@ -121,18 +132,13 @@ def all_modules() -> Iterable[HPIModule]:
 def _iter_my_roots() -> Iterable[Path]:
     import my  # doesn't import any code, because of namespace package
 
-    default_root = Path(__file__).absolute().parent.parent
-
-    try:
-        paths: List[str] = list(my.__path__._path)  # type: ignore[attr-defined]
-    except Exception as e:
-        logging.exception(e)
-        yield default_root
+    paths: List[str] = list(my.__path__)  # type: ignore[attr-defined]
+    if len(paths) == 0:
+        # should probably never happen?, if this code is running, it was imported
+        # because something was added to __path__ to match this name
+        raise RuntimeError("my.__path__ was empty, try re-installing HPI?")
     else:
-        if len(paths) == 0:
-            yield default_root
-        else:
-            yield from map(Path, paths)
+        yield from map(Path, paths)
 
 
 def _modules_under_root(my_root: Path) -> Iterable[HPIModule]:
@@ -206,8 +212,12 @@ def test_pure() -> None:
     """
     We want to keep this module clean of other HPI imports
     """
+    # this uses string concatenation here to prevent
+    # these tests from testing against themselves
     src = Path(__file__).read_text()
-    assert 'import '  + 'my' not in src
+    # 'import my' is allowed, but
+    # dont allow anything other HPI modules
+    assert re.findall('import ' + r'my\.\S+', src, re.M) == []
     assert 'from ' + 'my' not in src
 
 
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
Ah - suppose the test that checks that `import my` isnt in the source text failed since I'm trying to use the __path__

I don't imagine that could ever fail? The only way youre running this code is by importing it using something like `my.core.discovery_pure` or through HPI, which installs `my`.

Not sure if it should be removed or changed?

karlicoss:
Ah! Yeah I think making an exception for 'import my' makes sense

my/core/discovery_pure.py
```diff
@@ -114,10 +114,31 @@ def _extract_requirements(a: ast.Module) -> Requires:
...
+        yield from _modules_under_root(my_root)
+
+
+def _iter_my_roots() -> Iterable[Path]:
+    import my  # doesn't import any code, because of namespace package
+
+    default_root = Path(__file__).absolute().parent.parent
+
+    try:
+        paths: List[str] = list(my.__path__._path)  # type: ignore[attr-defined]
```
karlicoss, line 127: hm, I thought one could just use `list(pkg.__path__)`? What's _path for?

my/core/discovery_pure.py
```diff
@@ -114,10 +114,31 @@ def _extract_requirements(a: ast.Module) -> Requires:
...
+    except Exception as e:
+        logging.exception(e)
+        yield default_root
+    else:
+        if len(paths) == 0:
```
karlicoss, line 132: makes sense just in case, just curious if you know of this ever actually happening?

seanbreckenridge: I don't think it would ever be possible no

The answer to both of the questions is I was doing the same as you did in `config_ok`:

https://github.com/karlicoss/HPI/blob/master/my/core/__main__.py#L154-L159

my/core/discovery_pure.py
seanbreckenridge: Does seem to work the same, yeah

karlicoss:
Btw one think to keep in mind is that this will now return duplicate modules for files that exist in several overlays: 
And e.g. this would result in an arbitrary first one to be returned.
```
def module_by_name(name: str) -> HPIModule:
    for m in all_modules():
        if m.name == name:
            return m
    raise RuntimeError(f'No such module: {name}')
```

Guess ok for now, and can think of some way to disambiguate later...

seanbreckenridge:
>  And e.g. this would result in an arbitrary first one to be returned.

Is it arbitrary? Since we're iterating through `my.__path__`, which is the import order -- it should return the first item it finds which matches the name, which would be the first item in your `easy-install.pth`/import order, right?

Should still keep it in mind since `all_modules` can return duplicates, but `module_by_name` should resolve "correctly"

my/core/discovery_pure.py
karlicoss: Haha, if only I could remember now what I was thinking! Likely just knew less about packaging.
Read up https://docs.python.org/3/reference/import.html#module-path now
Seems that if it's a package (regardless whether it's namespace or ordinary), it always has __path__, but it may be empty. It only doesn't have __path__ if it's a 'leaf' module.
So guess `my` will always have it and can get rid of try-catch, but maybe makes sense to default (or throw?) if it's empty?

karlicoss:
Ahh, hmm good point! So maybe it's fine after all.

my/core/discovery_pure.py
seanbreckenridge: Alright, I'll remove the `_path` from here then

> or throw? if its empty?

Yeah, I guess I can remove the `default_path` altogether. If `__path__` is empty (how you'd have an empty `__path__` when youre running a file in the module I don't know), it wouldn't be able to compute anything in any case, and your installation is probably broken

karlicoss:
thanks!

Issue #148, opened 2021-03-28
Use click for arguments?


seanbreckenridge:
As mentioned in #138 , may want to switch to [`click`](https://click.palletsprojects.com/en/7.x/) for argument parsing, as its more composable and offers lots of features that the stdlib doesn't. Also handles printing [colored strings](https://click.palletsprojects.com/en/7.x/utils/#ansi-colors) if `colorama` is installed, prompt/confirming prompts in a nice way, along with other things like progress bars if you want to mess with that.

Since this is decorator based, also means a user has the possibility to extend the CLI by adding a hook somewhere

Once #138 has been merged, I'd be willing to write an initial PR for this, as I've been using `click` consistently for a few years now.

seanbreckenridge:
closing because of #155

PR #155, opened 2021-04-04
switch CLI to using click, handles printing colors

seanbreckenridge:
docs for click are good, would recommend skimming over them
https://click.palletsprojects.com/en/7.x/#documentation

everything is backwards-compatible with the previous interface, the only minor changes were to the `doctor` cmd which can now accept more than one item to run, and the `--skip-config-check` to skip the `config_ok` check if the user specifies to

Couldn't find anywhere in the `doc/` folder where I had to change, let me know if I missed something

added a test using `click.testing.CliRunner` (tests the CLI in an isolated environment), though additional tests which aren't testing the CLI itself (parsing arguments or decorator behaviour) can just call the functions themselves, as they no longer accept a argparser.Namespace and instead accept the direct arguments

I also replaced the `colorama.init` call in core.warnings, as click will make sure colorama is installed on windows if it needs it, and handles initializing it if needed, else it does typical ANSI-styled colors with [`click.style`](https://click.palletsprojects.com/en/7.x/api/#click.style). This **doesnt** use termcolor under the hood, it just does it manually (other than colorama on windows, click has no dependencies)

I know previously said that would PR this after writing the query interface, but didn't want to have to worry about making it compatible with argparse and click (and also, didn't feel like writing something in argparse, knowing I would re-write it almost immediately)

Does change the formatting for the CLI a bit (doesnt use epilog, uses the docstrings as descriptions), but the decorators and nesting is pretty readable. Any `click.group()` functions can then be used as a decorator themslelves, allowing you to nest like:

```python
@click.group():
def main():
	pass
	
@main.group()
def subgroup():
	pass

# options with '/' are interpreted as on/off flags
@subgroup.command()
@click.option("--verbose/--quiet", default=False, help="increase verbosity...")
def cmd(verbose: bool) -> None:
	"""docs for cmd go here"""
	click.echo("Hi!")
```

That would create an interface like `main.py subgroup cmd`, allowing you to add `click.options()` or `click.arguments()` along the way. Click supports lots of ways to deal with the complexities of parsing/sharing context between the groups and options, either adding callbacks, invoking one command from another, or adding a shared context between commands. [Advanced docs](https://click.palletsprojects.com/en/7.x/advanced/#invoking-other-commands)

Also, switched from using `print` to `click.echo` since it handles handles possible UnicodeErrors if the terminal is heavily misconfigured, strips ANSI colors if the output is a file. Overhead is tiny so I tend to use it by default so I don't have to worry about those problems in the future

the only major thing I think isn't immediately obvious is in click:

> Arguments work similarly to options but are positional. They also only support a subset of the features of options due to their syntactical nature. Click will also not attempt to document arguments for you and wants you to document them manually in order to avoid ugly help pages

So any documentation for arguments is done manually in the docstring

here are some examples of using the interface:

```
[ ~/Repos/HPI-to-master | click ] $ hpi module requires my.reddit
dependencies of my.reddit
'git+https://github.com/karlicoss/rexport'
'git+https://github.com/seanbreckenridge/pushshift_comment_export'
[ ~/Repos/HPI-to-master | click ] $ hpi module install --help
Usage: hpi module install [OPTIONS] MODULE

  Install dependencies for a module using pip

  MODULE is a specific module name (e.g. my.reddit)

Options:
  --user  same as pip --user
  --help  Show this message and exit.
```

```
[ ~/Repos/HPI-to-master | click ] $ hpi doctor --skip-config-check my.reddit my.github.all
✅ OK  : my.reddit
✅     - stats: {'saved': {'count': 51, 'last': datetime.datetime(2016, 9, 26, 0, 13, 19, tzinfo=<UTC>)}, 'comments': {'count': 4914}, 'submissions': {'count': 39}, 'upvoted': {'count': 923}}
✅ OK  : my.github.all
✅     - stats: {'events': {'events': {'count': 5697, 'last': datetime.datetime(2021, 4, 3, 7, 24, 37, tzinfo=<UTC>)}}, 'get_events': {'get_events': {'count': 5697, 'last': datetime.datetime(2021, 4, 3, 7, 24, 37, tzinfo=<UTC>)}}}
```

Only other thing I was worried about is adding a core dependency without a check?

If someone else is using this as an editable package and `git pulls` periodically to update this, and they didn't happen to have `click` installed (low chance given its popularity, but still), the `import click` would just fail

perhaps I should add something like?

```python
try:
	import click
except ModuleNotFoundError:
	print("this requires click for the user interface now.. run : python3 -m pip install --user click")
	sys.exit(1)
```

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -207,7 +207,7 @@ def _modules(*, all: bool=False) -> Iterable[HPIModule]:
         warning(f'Skipped {len(skipped)} modules: {skipped}. Pass --all if you want to see them.')
 
 
-def modules_check(verbose: bool, list_all: bool, quick: bool, for_modules: List[str]) -> None:
+def modules_check(*, verbose: bool, list_all: bool, quick: bool, for_modules: List[str]) -> None:
     if len(for_modules) > 0:
         # if you're checking specific modules, show errors
         # hopefully makes sense?
@@ -268,7 +268,7 @@ def modules_check(verbose: bool, list_all: bool, quick: bool, for_modules: List[
             info(f'    - stats: {res}')
 
 
-def list_modules(list_all: bool) -> None:
+def list_modules(*, list_all: bool) -> None:
     # todo add a --sort argument?
     tabulate_warnings()
 
@@ -309,14 +309,14 @@ def _requires(module: str) -> Sequence[str]:
     return r
 
 
-def module_requires(module: str) -> None:
+def module_requires(*, module: str) -> None:
     rs = [f"'{x}'" for x in _requires(module)]
     eprint(f'dependencies of {module}')
     for x in rs:
         click.echo(x)
 
 
-def module_install(user: bool, module: str) -> None:
+def module_install(*, user: bool, module: str) -> None:
     # TODO hmm. not sure how it's gonna work -- presumably people use different means of installing...
     # how do I install into the 'same' environment??
     import shlex
@@ -337,16 +337,21 @@ def main() -> None:
     Tool for HPI
     Work in progress, will be used for config management, troubleshooting & introspection
     '''
-    # note: code in groups doesnt run unless a subcommand is passed
+    # for potential future reference, if shared state needs to be added to groups
     # https://click.palletsprojects.com/en/7.x/commands/#group-invocation-without-command
     # https://click.palletsprojects.com/en/7.x/commands/#multi-command-chaining
+
+    # acts as a contextmanager of sorts - any subcommand will then run
+    # in something like /tmp/hpi_temp_dir
+    # to avoid importing relative modules by accident during development
+    # maybe can be removed later if theres more test coverage/confidence that nothing
+    # would happen?
+    import tempfile
 
     # use a particular directory instead of a random one, since
     # click being decorator based means its more complicated
     # to run things at the end (would need to use a callback or pass context)
     # https://click.palletsprojects.com/en/7.x/commands/#nested-handling-and-contexts
-
-    import tempfile
 
     tdir: str = os.path.join(tempfile.gettempdir(), 'hpi_temp_dir')
     if not os.path.exists(tdir):
@@ -370,7 +375,7 @@ def doctor_cmd(verbose: bool, list_all: bool, quick: bool, skip_conf: bool, modu
     if not skip_conf:
         config_ok()
     # TODO check that it finds private modules too?
-    modules_check(verbose, list_all, quick, list(module))
+    modules_check(verbose=verbose, list_all=list_all, quick=quick, for_modules=list(module))
 
 
 @main.group(name='config', short_help='work with configuration')
@@ -396,7 +401,7 @@ def config_create_cmd() -> None:
 @click.option('--all', 'list_all', is_flag=True, help='List all modules, including disabled')
 def module_cmd(list_all: bool) -> None:
     '''List available modules'''
-    list_modules(list_all)
+    list_modules(list_all=list_all)
 
 
 @main.group(name='module', short_help='module management')
@@ -413,7 +418,7 @@ def module_requires_cmd(module: str) -> None:
 
     MODULE is a specific module name (e.g. my.reddit)
     '''
-    module_requires(module)
+    module_requires(module=module)
 
 
 @module_grp.command(name='install', short_help='install module deps')
@@ -426,7 +431,7 @@ def module_install_cmd(user: bool, module: str) -> None:
     MODULE is a specific module name (e.g. my.reddit)
     '''
     # todo could add functions to check specific module etc..
-    module_install(user, module)
+    module_install(user=user, module=module)
 
 
 # todo: add more tests?
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -437,11 +437,10 @@ def module_install_cmd(user: bool, module: str) -> None:
 
 
 def test_doctor_single():
-    # cli test
     from click.testing import CliRunner
-    result = CliRunner().invoke(main, ['doctor', '--skip-config-check', 'my.core'])
+    result = CliRunner().invoke(main, ['module', 'requires', 'my.github.ghexport'])
     assert result.exit_code == 0
-    assert "no 'stats' function" in result.output
+    assert "github.com/karlicoss/ghexport" in result.output
 
 
 if __name__ == '__main__':
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -436,7 +436,7 @@ def module_install_cmd(user: bool, module: str) -> None:
 # other python code
 
 
-def test_doctor_single():
+def test_requires() -> None:
     from click.testing import CliRunner
     result = CliRunner().invoke(main, ['module', 'requires', 'my.github.ghexport'])
     assert result.exit_code == 0
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -1,15 +1,13 @@
 import functools
 import importlib
 import os
+import sys
+import traceback
+from typing import Optional, Sequence, Iterable, List
 from pathlib import Path
 from subprocess import check_call, run, PIPE, CompletedProcess
-import sys
-from typing import Optional, Sequence, Iterable, List
-import traceback
 
-from . import LazyLogger
-
-log = LazyLogger('HPI cli')
+import click
 
 
 @functools.lru_cache()
@@ -58,8 +56,12 @@ def run_mypy(pkg: ModuleType) -> Optional[CompletedProcess]:
     return mres
 
 
+# use click.echo over print since it handles handles possible Unicode errors,
+# strips colors if the output is a file
+# https://click.palletsprojects.com/en/7.x/quickstart/#echoing
 def eprint(x: str) -> None:
-    print(x, file=sys.stderr)
+    # err=True prints to stderr
+    click.echo(x, err=True)
 
 def indent(x: str) -> str:
     return ''.join('   ' + l for l in x.splitlines(keepends=True))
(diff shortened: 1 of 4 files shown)


seanbreckenridge:
Would resolve #148

seanbreckenridge:
Hmm, had a feeling this might fail in the CI...

seems the guess_stats returned a module instead of the stats for `my.core`...

```
______________________________ test_doctor_single ______________________________

    def test_doctor_single():
        # cli test
        from click.testing import CliRunner
        result = CliRunner().invoke(main, ['doctor', '--skip-config-check', 'my.core'])
        assert result.exit_code == 0
>       assert "no 'stats' function" in result.output
E       assert "no 'stats' function" in '✅ OK  : my.core                                           \n❗      - stats:                      computing failed\n  ...core/__main__.py", line 261, in modules_check\n       res = stats()\n   TypeError: \'module\' object is not callable\n'
E        +  where '✅ OK  : my.core                                           \n❗      - stats:                      computing failed\n  ...core/__main__.py", line 261, in modules_check\n       res = stats()\n   TypeError: \'module\' object is not callable\n' = <Result okay>.output

/home/runner/work/HPI/HPI/my/core/__main__.py:444: AssertionError
```

seanbreckenridge:
switched it to testing 'module requires' since that uses 'discovery_pure' and shouldn't fail due to something in the CI since its parsing the AST instead

Just wanted to have an example of using the click testrunner in there to show it was possible

my/core/__main__.py
```diff
@@ -360,69 +329,121 @@ def module_install(args: Namespace) -> None:
...
+
+# todo: add more tests?
+# its standard click practice to have the function click calls be a separate
+# function from the decorated function, as it allows the application-specific code to be
+# more testable. also allows hpi commands to be imported and called manually from
+# other python code
+
+
+def test_requires() -> None:
+    from click.testing import CliRunner
```
karlicoss, line 440: > added a test using click.testing.CliRunner (tests the CLI in an isolated environment)

Neat! Although I've started wondering how it could possibly work since something like cgroups isn't portable and generally would need root permissions etc -- seems that it's pretty basic, just hacks the env variables & IO streams
https://github.com/pallets/click/blob/c4542263a96dd067ed1ded12a83553da05bb61dd/src/click/testing.py#L172

And the filesystem isolation is merely a chdir https://github.com/pallets/click/blob/c4542263a96dd067ed1ded12a83553da05bb61dd/src/click/testing.py#L384 Also isn't running by default, which is weird, arguably one of the most important reasons to isolate? 

I might contribute to their docs, feel like it can be misleading if people don't look in the source.
But otherwise still super convenient if one is aware of the limitations.

my/core/__main__.py
```diff
@@ -360,69 +329,121 @@ def module_install(args: Namespace) -> None:
...
+    # note: code in groups doesnt run unless a subcommand is passed
+    # https://click.palletsprojects.com/en/7.x/commands/#group-invocation-without-command
+    # https://click.palletsprojects.com/en/7.x/commands/#multi-command-chaining
 
-    cp = sp.add_parser('config', help='Work with configuration')
-    scp = cp.add_subparsers(dest='config_mode')
-    if True:
-        ccp = scp.add_parser('check', help='Check config')
-        ccp.set_defaults(func=config_check_cli)
+    # use a particular directory instead of a random one, since
```
karlicoss, line 344: Ah fair enough -- and anyway being a `main` it basically means it's a context manager of sorts on its own, so chdir won't hurt. Maybe the only place where it can leave process in a different state is in tests?

my/core/__main__.py
```diff
@@ -360,69 +329,121 @@ def module_install(args: Namespace) -> None:
...
-    dp = sp.add_parser('doctor', help='Run various checks')
-    dp.add_argument('--verbose', action='store_true', help='Print more diagnosic infomration')
-    dp.add_argument('--all'    , action='store_true', help='List all modules, including disabled')
-    dp.add_argument('--quick'  , action='store_true', help='Only run partial checks (first 100 items)')
-    dp.add_argument('module', nargs='?', type=str   , help='Pass to check a specific module')
-    dp.set_defaults(func=doctor)
+    Tool for HPI
+    Work in progress, will be used for config management, troubleshooting & introspection
+    '''
+    # note: code in groups doesnt run unless a subcommand is passed
```
karlicoss, line 340: Hmm, what does it imply? As in, isn't that what one would expect? 
As far as I can tell from the docs it basically means it lets to get rid of things like `print_help` handling (nice!)? And I guess often there is a pattern in argparse where you do `if/elif` switching on `args.subcommand` or something like that, which also isn't a problem in click?

my/core/__main__.py
```diff
@@ -360,69 +329,121 @@ def module_install(args: Namespace) -> None:
...
-    ops = op.add_subparsers(dest='module_mode')
-    if True:
-        add_module_arg = lambda x: x.add_argument('module', type=str, help='Module name (e.g. my.reddit)')
 
-        opsr = ops.add_parser('requires', help='Print module requirements')
-        # todo not sure, might be worth exposing outside...
-        add_module_arg(opsr)
-        opsr.set_defaults(func=module_requires)
+@main.command(name='doctor', short_help='run various checks')
+@click.option('--verbose/--quiet', default=False, help='Print more diagnostic information')
```
karlicoss, line 358: >  options with '/' are interpreted as on/off flags

uhh neat -- didn't know

karlicoss (commented):
nice, looks great, thanks! 
gonna check this out in cli as well just in case

my/core/__main__.py
seanbreckenridge: Yeah, I was considering adding a contextmangaer so it'd be possible to do that in the commands itself, when it was necessary, but I didn't know what it was needed for. Is it the `importlib` statements?

my/core/__main__.py
seanbreckenridge: Ah, yeah -- previously I was using `tempfile.mkdtemp` which ended up creating a bunch of temporary directories in `/tmp`, so it was sort of a warning. I just left it there and added the links incase you wanted to add some global flags/context later on

my/core/__main__.py
```diff
@@ -226,12 +207,11 @@ def _modules(*, all: bool=False) -> Iterable[HPIModule]:
         warning(f'Skipped {len(skipped)} modules: {skipped}. Pass --all if you want to see them.')
 
 
-def modules_check(args: Namespace) -> None:
-    verbose: bool         = args.verbose
-    quick:   bool         = args.quick
-    module: Optional[str] = args.module
-    if module is not None:
-        verbose = True # hopefully makes sense?
+def modules_check(verbose: bool, list_all: bool, quick: bool, for_modules: List[str]) -> None:
```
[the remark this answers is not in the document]
seanbreckenridge: Yeah, that makes sense -- I might just give these defaults instead, so that they're easier to use if imported by the user

my/core/__main__.py
seanbreckenridge: Often described in the docs like `--save/--no-save` in the docs; is why that doesnt have the `is_flag=True` option

my/core/__main__.py
karlicoss: Yeah, I think likely to avoid importing stuff by accident from cwd during development, with various sorts of automatic discoveries can be confusing. More of a paranoid thing I guess, maybe can remove later with more confidence/test coverage? Can be confusing too if some commands start producing outputs at some point, and people use relative paths

my/core/__main__.py
seanbreckenridge: Tangentially related, you can choose to run a group if a subcommand isnt passed:

```
@click.group(invoke_without_command=True)
@click.pass_context
def cli(ctx):
    if ctx.invoked_subcommand is None:
        click.echo('I was invoked without subcommand')
    else:
        click.echo('I am about to invoke %s' % ctx.invoked_subcommand)

@cli.command()
def sync():
    click.echo('The subcommand')
```

If a subcommand is passed, it runs the function body, then invokes the subcommand

my/core/__main__.py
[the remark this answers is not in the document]
karlicoss: I guess the only problem is that it's hard to keep it consistent between click defaults and function defaults?

seanbreckenridge:
Also, not sure if the `--user` problem has been an issue for you recently. I remember it used to just break since it didn't have permission on my mac, has always worked on my linux machine as far as I remember

Probably 2 or 3 years ago there was a pip update that fixed the issue on the mac, where if it wasn't able to install into the python location for the globally installed python (typically that would need `sudo`) it would switch to behavior similar to the `--user` flag.

Since then things like `python3 -m pip install packageName` have worked fine on mac for me, they just send a warning sayings its switching to the user-specific directory

Fine to leave it there for the foreseeable future, as who knows what version of pip someone is on, but as a data point, hasn't been a problem for me for a while.

my/core/__main__.py
seanbreckenridge: Yeah, thats a good point. Typically the defaults for click are pretty good - the only time its tripped me up is tuples vs. lists

my/core/__main__.py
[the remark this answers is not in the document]
seanbreckenridge: Yeah, thats a good point. Typically the defaults for click are pretty good - the only time its tripped me up is tuples vs. lists

karlicoss:
Hmmm indeed, somehow it works now! 
But still looks like it's not 100% consistent :thinking:  

```
$ which pip3
/home/karlicos/.local/bin/pip3

$ pip3 install httpx && pip3 uninstall --yes httpx
Defaulting to user installation because normal site-packages is not writeable
Collecting httpx
  Using cached httpx-0.17.1-py3-none-any.whl (68 kB)
Installing collected packages: httpx
Successfully installed httpx-0.17.1
WARNING: You are using pip version 20.2.2; however, version 21.0.1 is available.
You should consider upgrading via the '/usr/bin/python3.8 -m pip install --upgrade pip' command.
...

$ /usr/bin/pip3 install httpx && pip3 uninstall --yes httpx
Collecting httpx
  Using cached httpx-0.17.1-py3-none-any.whl (68 kB)
Installing collected packages: httpx
Successfully installed httpx-0.17.1
...
```

So somehow both succeed, but only user pip is printing the warning `Defaulting to user installation because normal site-packages is not writeable`? Also interesting that only user pip is printing upgrade warning, even though both are on the old version.

PR #157, opened 2021-04-05
add hpi query command

### LINKED ISSUE #138: Add core.query/serialize?

seanbreckenridge:
Resolves #138 

had to restructure `query.py` code a bit to allow `query_range.py`, which supports filtering items by coercing them to some shared type, used in the cli like:

`hpi query --order-type datetime --after '2016-01-01 00:00:00' --before '2019-01-01 00:00:00' my.reddit.comments`

Can do fun stuff in bash like

```bash
$ hpi query --recent 4w my.github.ghexport.events | jq -r '.[] | .link'\
 | sort | grep -vx null | cut -d'/' -f-5 | sort |\
 uniq -c | sort -n | tail | tac
     99 https://github.com/karlicoss/HPI
     20 https://github.com/seanbreckenridge/HPI
     11 https://github.com/seanbreckenridge/mpv-history-daemon
      7 https://github.com/seanbreckenridge/autotui
      6 https://github.com/seanbreckenridge/mint
      6 https://github.com/seanbreckenridge/HPI-to-master
      5 https://github.com/seanbreckenridge/glue
      4 https://github.com/seanbreckenridge/lolexport
      4 https://github.com/seanbreckenridge/aw-watcher-window
      4 https://github.com/karlicoss/arctee
```

Or just the more obvious, dump the results from all of this to JSON (using `my.core.serialize`):

```
$ hpi query my.browsing.history > dump.json
$ hpi query my.github.events > dump2.json
 $ du -h dump*.json
67M     dump2.json
89M     dump.json
```

Or drop me into the repl with the results of this query:

```
 $ hpi query --recent 1d my.browsing.history -o repl

Interact with the results by using the res variable

Python 3.9.2 (default, Feb 20 2021, 18:40:11)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: len(res)
Out[1]: 205
```

```
$ hpi query --help
Usage: hpi query [OPTIONS] FUNCTION_NAME...

  This allows you to query the results from one or more functions in HPI

  By default this runs with '-o json', converting the results to JSON and
  printing them to STDOUT

  You can specify '-o pprint' to just print the objects using their repr, or
  '-o repl' to drop into a ipython shell with access to the results

  While filtering using --order-key datetime, the --after, --before and
  --within flags parse the input to their datetime and timedelta
  equivalents. datetimes can be epoch time, the string 'now', or an date
  formatted in the ISO format. timedelta (durations) are parsed from a
  similar format to the GNU 'sleep' command, e.g. 1w2d8h5m20s -> 1 week, 2
  days, 8 hours, 5 minutes, 20 seconds

  As an example, to query reddit comments I've made in the last month

  hpi query --order-type datetime --before now --within 4w my.reddit.comments
  or...
  hpi query --recent 4w my.reddit.comments

  Can also query within a range. To filter comments between 2016 and 2018:
  hpi query --order-type datetime --after '2016-01-01 00:00:00' --before '2019-01-01 00:00:00' my.reddit.comments

Options:
  -o, --output [json|pprint|repl]
                                  what to do with the result [default: json]
  -k, --order-key TEXT            order by an object attribute or dict key on
                                  the individual objects returned by the HPI
                                  function

  -t, --order-type [datetime|int|float|none]
                                  order by searching for some type on the
                                  iterable

  --after TEXT                    while ordering, filter items for the key or
                                  type larger than or equal to this

  --before TEXT                   while ordering, filter items for the key or
                                  type smaller than this

  --within TEXT                   a range 'after' or 'before' to filter items
                                  by. see above for further explanation

  --recent TEXT                   a shorthand for '--order-type datetime
                                  --reverse --before now --within'. e.g.
                                  --recent 5d

  --reverse / --no-reverse        reverse the results returned from the
                                  functions

  --limit INTEGER                 limit the number of items returned from the
                                  (functions)

  --drop-unsorted                 If the order an item can't be determined
                                  while ordering, drop those items from the
                                  results

  --wrap-unsorted                 If the order an item can't be determined
                                  while ordering, drop those items from the
                                  results

  --raise-exceptions              If any errors are returned (as objects, not
                                  raised) from the functions, raise them

  --drop-exceptions               Ignore any errors returned as objects from
                                  the functions

  --help                          Show this message and exit.
```


my/core/__main__.py
```diff
@@ -329,6 +329,65 @@ def module_install(*, user: bool, module: str) -> None:
...
+    elif output == 'pprint':
+        from pprint import pprint
+
+        pprint(res)
+    else:
+        # output == 'repl'
+        try:
+            import IPython  # type: ignore[import]
+        except ModuleNotFoundError:
+            eprint("'repl' requires ipython, install it with 'python3 -m pip install ipython'")
```
karlicoss, line 384: could fallback on normal python, IIRC it's possible to embed it as well via https://docs.python.org/3/library/code.html#code.interact
but I guess not a big deal

my/core/__main__.py
```diff
@@ -434,6 +493,145 @@ def module_install_cmd(user: bool, module: str) -> None:
...
+@main.command(name='query', short_help='query the results of a HPI function')
+@click.option('-o',
+              '--output',
+              default='json',
+              type=click.Choice(['json', 'pprint', 'repl']),
+              help='what to do with the result [default: json]')
+@click.option('-k',
+              '--order-key',
+              default=None,
+              type=click.STRING,
```
karlicoss, line 505: `type=str` seems equivalent according to the docs?

my/core/query.py
```diff
@@ -206,6 +238,18 @@ def _drop_unsorted(itr: Iterator[ET], orderfunc: OrderFunc) -> Iterator[ET]:
         yield o
 
 
+# try getting the first value from the iterator
+# similar to my.core.common.warn_if_empty? this doesnt go through the whole iterator though
```
karlicoss, line 242: oh good point, maybe I forgot about more_itetrools back then
although it's injected in the iterable itself, and the only way to use the iterable is to iterate all of it.. so I guess the usecase is a bit different

my/core/query_range.py
```diff
@@ -0,0 +1,526 @@
...
+from .query import (
+    QueryException,
+    select,
+    OrderFunc,
+    Where,
+    _handle_generate_order_by,
+    ET,
+)
+
+fromisoformat: Callable[[str], datetime]
```
karlicoss, line 27: maybe this snippet belongs to `core/compat`? Could remove the duplicate from `common` as well

my/core/query_range.py
```diff
@@ -0,0 +1,526 @@
...
+    # cant specify all three
+    with pytest.raises(QueryException, match=r"Cannot specify 'after', 'before' and 'within'"):
+        dt_parse_range(unparsed_range=RangeTuple(str(start_date), str(end_date.timestamp()), "7d"))
+
+    # if you specify noting, should return None
+    res3 = dt_parse_range(unparsed_range=RangeTuple(None, None, None))
+    assert res3 is None
+
+
+def test_parse_timedelta_string() -> None:
```
karlicoss, line 496: would be nice in the doctest, but IIRC it's not running at the moment, had some weird issues setting up for HPI

my/core/query_range.py
```diff
@@ -0,0 +1,526 @@
...
+
+    res = parse_timedelta_string("1w5d5h10m50s")
+    assert res == timedelta(days=7.0 + 5.0, hours=5.0, minutes=10.0, seconds=50.0)
+
+
+def test_parse_datetime_float() -> None:
+
+    pnow = parse_datetime_float("now")
+    sec_diff = abs((pnow - datetime.now().timestamp()))
+    # should probably never fail? could mock time.time
```
karlicoss, line 511: eh.. IMO good enough, mocking time seems a bit insane  xD

my/core/query_range.py
```diff
@@ -0,0 +1,526 @@
...
+    except ValueError:
+        pass
+    else:
+        return ds_float
+    try:
+        # isoformat - default format when you call str() on datetime
+        return fromisoformat(ds).timestamp()
+    except ValueError:
+        raise QueryException(f"Was not able to parse {ds} into a datetime")
+    # todo: add isoparse from my.core.common? not sure how it handles
```
karlicoss, line 86: fromisoformat should support timezones.. although it wants a UTC offset like `+04:00`

karlicoss (commented):
looks good, thanks!

my/core/query_range.py
seanbreckenridge: yeah, didnt add it as a doctest as I've had similar issues trying to get those to work

seanbreckenridge:
Just realized I didn't have `date` as an option for the `--order-type`

most items in HPI have `datetime` but since it all works with `date` and its easy to add, might as well give the option

my/core/query_range.py
seanbreckenridge: yeah, didnt want to add the extra dependency either, there are some third part libs that can handle this

Issue #64, opened 2020-07-11
ActivityWatch module?


alexriabtsev:
Is there any way to import [ActivityWatch](https://github.com/ActivityWatch/activitywatch) info?
They have API with docs - https://docs.activitywatch.net/en/latest/rest.html

seanbreckenridge:
I've played around with this a bit, and the docs don't really document everything that well (seems they've been in the process of migrating to rust-based API as well). had to go and poke in [here](https://github.com/ActivityWatch/aw-server/blob/master/aw_server/rest.py#L112-L139), and couldn't get the delete route to work after fiddling with it for a bit. I'm not sure how this would work with multiple buckets/generic events either. The window watcher/afk modules are standard, but how would one support loading in arbitrary data buckets?

As it says there:

```
The API is currently under development, and is subject to change. It will be documented in better detail when first version has been frozen.
```

I really did like `aw-window-watcher` though, so I ended up [forking that](https://github.com/seanbreckenridge/aw-watcher-window) to log to a CSV file instead, and have a ddata loader for that in [`my.window_watcher`](https://github.com/seanbreckenridge/HPI/blob/master/my/window_watcher.py)

Generally curious if you're using ActivityWatcher for anything else, is there some data you'd want to expose through the API?

madelinecameron:
:wave: I wrote a basic interface for ActivityWatch. 

My extractor & DAL code is here: https://github.com/hpi/activitywatch

My synchronization code is https://github.com/hpi/aw-sync-client and https://github.com/hpi/aw-sync-server

It works alright from what I've seen but there could still be fiddling needed. The way I accomplished this is using the query endpoints then synchronizing to an off-site server (not at all required tbh, I should modify the extractor to call directly), then my extractor (currently) calls out to my external AW sync server. 

I did this so I can run AW on my personal Linux & WIndows along with any work computers.

krillin666:
> wave I wrote a basic interface for ActivityWatch.
> 
> My extractor & DAL code is here: [hpi/activitywatch](https://github.com/hpi/activitywatch)
> 
> My synchronization code is [hpi/aw-sync-client](https://github.com/hpi/aw-sync-client) and [hpi/aw-sync-server](https://github.com/hpi/aw-sync-server)
> 
[...]

Issue #102, opened 2020-10-26
Discussion: extending base HPI/overlays/overrides


karlicoss:
Related issues: https://github.com/karlicoss/HPI/issues/12, https://github.com/karlicoss/HPI/issues/46; but I think worth a separate discussion.

From my experience, it's pretty hard to predict how other people want to use their data:

- you might miss some attributes they care about
- some people want to be more paranoid or more defensive (e.g. timezone handling/None safety/etc)
- they might want to do some extra filtering
- they might want to merge in extra data sources or suppress existing

The list is endless! So it would be nice if it was possible to easily override small bits of HPI modules/mix in other data, etc.

The main goals are:

- low effort
- good interop, e.g. ability to keep with the upstream, use modules coming from separate repositories, etc.
- ideally mypy friendly. This kind of means 'not too dynamic and magical', which is ultimately a good thing even if you don't care about mypy.

Once again, I see Emacs as a good role model. Everything is really decentralized, you have some core library, you have certain patterns that everyone follows... but apart from that the modules are mostly independent.
Many people still use 'monolith' base configurations (e.g. Doom/Spacemacs), because it's kinda convenient, as long as you have a maintainer. Arguably this is what this repository is at the moment, although it's obviously not as popular as Emacs distributions. 

Emacs fits these goals well:

- low effort: the simplest way to confugure something is to override a variable in your config (thanks to dynamic scope, it 'just works'). 

  You can even literally override whole functions as a means of quickly getting the behaviour you want.

- good interop: yes, unless the developer broke some APIs, usually you can safely update the upstream module.

---------------------

How to achieve this within HPI:

For combining independent modules together (say, something like `my/youtube.py` and `my/vimeo.py` coming from different repositories), the easiest is to use:

- symlinks (at least if you have just a few files/directories to mixin)
- namespace packages (more on them later)

Now, the tricky case is when you want to partially override something. 
The first option is: fork & apply your modifications on top. For example: https://github.com/seanbreckenridge/HPI

   - effort: very straightforward 
   - interop: merging with the upstream a bit manual, but if you use atomic commits & interactive rebase/cherry pick, should be manageable
   - at least not any more magical than the original repository

Not sure if there is much to discuss here, so straight to the second and a more flexible option. 

Once again, we rely on namespace packages! I'll just explain on a couple of examples, since it's easier.

- example: mixing in a data source

[...]

PR #158, opened 2021-04-07
Add a few more `my` interfaces

madelinecameron:
Addresses https://github.com/karlicoss/HPI/issues/64

- Withings
- Level Health
- Eight Sleep
- ActivityWatch (local-only RescueTime, also write my own synchronization code so you can sync multiple computer's AW data)
- Typeform
- Todoist

Unfortunately the extractors are written in node then I wrote DALs for each in Python. Maybe I am being presumptive but I think exposing the idea that HPI doesn't have to be solely in python is fun. Love being able to leverage Python's data library ecosystem though, much better than node for this :joy:. 

 The `my` interfaces are pretty barebones and may not be up to standards, so happy to modify if there is a certain standard or direction you like better.

### DIFF
--- a/my/eightsleep.py
+++ b/my/eightsleep.py
@@ -40,26 +40,26 @@ def dataframe(defensive: bool=True) -> DataFrameT:
         for s in sessions():
             try:
                 d = {
-                    'ts': s['ts'],
-                    'score'      : s['score'],
-                    'stages': pd.DataFrame(s['stages']),
-                    'tossAndTurns': pd.DataFrame(s['tossAndTurns']),
-                    'tempRoomC': pd.DataFrame(s['tempRoomC']),
-                    'tempBedC': pd.DataFrame(s['tempBedC']),
-                    'respiratoryRate': pd.DataFrame(s['respiratoryRate']),
-                    'heartRate': pd.DataFrame(s['heartRate']),
-                    'hrv': pd.DataFrame(s['hrv']),
-                    'rmssd': pd.DataFrame(s['rmssd']),
-                    'stages': pd.DataFrame(s['stages']),
-                    'presenceDuration': s['presenceDuration'] if 'presenceDuration' in s else 0,
-                    'sleepDuration': s['sleepDuration'] if 'sleepDuration' in s else 0,
-                    'deepPercent': s['deepPercent'] if 'deepPercent' in s else 0,
-                    'presenceStart': s['presenceStart'] if 'presenceStart' in s else 0,
-                    'presenceEnd': s['presenceEnd'] if 'presenceEnd' in s else 0,
-                    'sleepStart': s['sleepStart'] if 'sleepStart' in s else 0,
-                    'sleepEnd': s['sleepEnd'] if 'sleepEnd' in s else 0,
+                    'ts'               : s['ts'],
+                    'score'            : s['score'],
+                    'stages'           : pd.DataFrame(s['stages']),
+                    'tossAndTurns'     : pd.DataFrame(s['tossAndTurns']),
+                    'tempRoomC'        : pd.DataFrame(s['tempRoomC']),
+                    'tempBedC'         : pd.DataFrame(s['tempBedC']),
+                    'respiratoryRate'  : pd.DataFrame(s['respiratoryRate']),
+                    'heartRate'        : pd.DataFrame(s['heartRate']),
+                    'hrv'              : pd.DataFrame(s['hrv']),
+                    'rmssd'            : pd.DataFrame(s['rmssd']),
+                    'stages'           : pd.DataFrame(s['stages']),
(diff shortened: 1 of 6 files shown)


seanbreckenridge:
Ah, I didn't realize you weren't watching [this repo](https://github.com/hpi/hpi), so you may have not seen the issue I raised [here](https://github.com/hpi/hpi/issues/1)

To expand on it a bit;

I used to maintain a direct fork/merge commits from this repo back into my own copy of HPI [here](https://github.com/seanbreckenridge/HPI), but that became more and more difficult to as time went on...

The solution for me was in the [SETUP.org](https://github.com/karlicoss/HPI/blob/master/doc/SETUP.org#addingmodifying-modules) file all along (described as the 'proper way')

I've since stopped merging commits that, and now install [both packages as editable](https://github.com/seanbreckenridge/HPI/blob/e618db9007a5bf6e2d0da2d459c956e04c1388ac/install#L42-L69), which means I keep up to date with this repository and can still add/overlay modules on top of that

i.e. you can have multiple `my/` directories installed at the same time, if they're installed as namespace packages (relevant lines in setup.py [here](https://github.com/karlicoss/HPI/blob/fb4924300508ce16accc3b79bd7253d127872712/setup.py#L17) and [here](https://github.com/seanbreckenridge/HPI/blob/master/setup.py#L4-L8)):

```
 $ python3
In [1]: from my.coding.commits import commits  # something from karlicoss/HPI

In [2]: from my.zsh import history  # something from my own repo
```

Another example of an overlay [here](https://github.com/karlicoss/hpi-personal-overlay)

Not saying these items can't be merged into here, thats up to karicoss, I've just found HPI to be more useful as a configuration/caching layer, giving me all the tools to load my data, but everyone uses different services/data sources, so I'd rather keep modules on my own branch. Otherwise, we'd eventually have hundreds of data sources in this one repository, and I don't know if anyone wants to maintain that :sweat_smile:

my/levelshealth.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def inputs() -> Sequence[Path]:
+    return get_files(levelshealth.export_path)
+
+
+import levelshealth.dal as dal
+
+
+def glucoseScores():
```
karlicoss, line 30: if you annotate return types with `Iterator` (even `Iterator[Any]` should do), then `hpi doctor my.levelshealth` would discover them (so you won't need manual `stats` function)

my/levelshealth.py
```diff
@@ -0,0 +1,41 @@
+'''
+Levels Health data
```
karlicoss, line 2: would be nice to include the link to the service in the docstring as well

my/levelshealth.py
```diff
@@ -0,0 +1,41 @@
...
+
+REQUIRES = [
+    'git+https://github.com/hpi/levelshealth',
+]
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Sequence, Iterable
+
+from .core import Paths, get_files
```
karlicoss, line 13: we've recently been trying to start switching module code to fully qualified imports (`from my.core import...`) -- the reason is that eventually core will be split as a separate package, and then relative imports in modules won't make much sense
https://github.com/karlicoss/HPI/issues/102#issuecomment-796476003

my/eightsleep.py
```diff
@@ -0,0 +1,82 @@
...
+                }
+            except Exception as e:
+                # TODO use the trace? otherwise str() might be too short..
+                # r.g. for KeyError it only shows the missing key
+                # todo check for 'defensive'
+                d = {'error': f'{e} {s}'}
+            yield d
+    import pandas as pd # type: ignore
+    df = pd.DataFrame(it())
+    print(df)
```
karlicoss, line 72: best to avoid printing to stdout unless necessary

karlicoss:
> Unfortunately the extractors are written in node then I wrote DALs for each in Python

No that's definitely cool, and I'm definitely pro using whatever tools people can use! 

> Not saying these items can't be merged into here, thats up to karlicoss, I've just found HPI to be more useful as a configuration/caching layer, giving me all the tools to load my data, but everyone uses different services/data sources, so I'd 
rather keep modules on my own branch. 

Yep, definitely suggest trying this approach out too!

> Otherwise, we'd eventually have hundreds of data sources in this one repository, and I don't know if anyone wants to maintain that 

Having too many modules is definitely an issue -- I dread the fate of the Spacemacs maintainer! I think for now I'm basically happy to merge things, stuff doesn't seem to break much; but need to start thinking of ways to making it more manageable in advance perhaps.

Although I would prefer if they had tests/ example data pointers somewhere, since otherwise I can't guarantee I won't break it :)

my/eightsleep.py
madelinecameron: Ahhhh! Thanks! Didn't meant to leave this in.

Issue #147, opened 2021-03-28
Fix spelling mistake for my.coding.commits


seanbreckenridge:
Has been mentioned in [other issues](https://github.com/karlicoss/HPI/pull/76#issue-480684785), but thought I'd just create an issue to track this:

In lots of places in: https://github.com/karlicoss/HPI/blob/master/my/coding/commits.py#L60

`commited` should be spelled as `committed`

Comments from https://github.com/karlicoss/HPI/pull/76#issuecomment-687883217

(me)

> Perhaps a `@property` wrapper?

(karlicoss)

> Yep, I think this is the way to go with such things. This particular module is probably used by very few people, but I guess best not to break it unless absolutely necessary.

> Also a deprecation warning would be nice, I was thinking of using something like tantale/deprecated library, or just vendorizing a small, simple decorator to avoid an extra core dependency.

karlicoss:
ah! yeah, a common typo I do, thanks for letting me know 😅
gonna add to the spaced repetition queue

PR #159, opened 2021-04-07
my.coding.commits: fix misspelling/add warning

seanbreckenridge:
should fix https://github.com/karlicoss/HPI/issues/147, unless you wanted to vendorize/add something like [tantale/deprecated](https://github.com/tantale/deprecated)

personally think sending the warnings manually is fine


karlicoss:
thanks!

PR #164, opened 2021-04-19
cli/query: bugfix where datetime was ignored

seanbreckenridge:
missed an elif by mistake, realized
this was an issue when I tried
the example in the --help text
and it failed

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -672,7 +672,7 @@ def query_cmd(
     chosen_order_type: Optional[Type]
     if order_type == "datetime":
         chosen_order_type = datetime
-    if order_type == "date":
+    elif order_type == "date":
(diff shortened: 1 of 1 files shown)


karlicoss:
Thanks! I often use an ad-hoc dict to simulate exhaustive pattern matching -- makes it harder to make a mistake :) like
```
chosen_order_type = {
    "datetime": datetime,
    "int"     : int,
    "date"    : date,
}.get(order_type)
```

seanbreckenridge:
Yeah, I've done that in the past as well, feels a bit clunky creating a dictionary every time but I guess performance isnt really a concern as python objects are just dictionaries all the way down

wish the switch-case statement could be backported to earlier versions of python, solves this situation pretty nicely

PR #169, opened 2021-04-28
cli/query: add --stream flag

seanbreckenridge:
allows you to do something like

`hpi query --stream my.reddit.comments`

to stream the JSON objects one per line, makes
it nicer to pipe into 'jq'/'fzf' instead
of having to process the giant list
at the end

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -410,6 +410,7 @@ def _locate_functions_or_prompt(qualified_names: List[str], prompt: bool = True)
 def query_hpi_functions(
     *,
     output: str = 'json',
+    stream_json: bool = False,
     qualified_names: List[str],
     order_key: Optional[str],
     order_by_value_type: Optional[Type],
@@ -431,7 +432,7 @@ def query_hpi_functions(
     # chain list of functions from user, in the order they wrote them on the CLI
     input_src = chain(*(f() for f in _locate_functions_or_prompt(qualified_names)))
 
-    res = list(select_range(
+    res = select_range(
         input_src,
         order_key=order_key,
         order_by_value_type=order_by_value_type,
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
e.g.

The sort call is blocking here but `jq` is able to process while `hpi query` is still running:

```
hpi query -s my.browsing.history | jq '.url' | sort | uniq -c | sort -n | tail
   1427 "https://discord.com/channels/267799767843602452/268515018897620995"
   1610 "https://discord.com/channels/460491088004907029/461199124205797439"
   2010 "https://discord.com/channels/267799767843602452/285862443689967617"
   2065 "https://discord.com/channels/267799767843602452/378707055327969300"
   2090 "https://discord.com/channels/267799767843602452/375914227371606016"
   2386 "https://discord.com/channels/267799767843602452/457596715722211349"
   3123 "https://github.com/seanbreckenridge"
   3898 "https://www.youtube.com/feed/subscriptions"
   5770 "https://discord.com/channels/267799767843602452/285547286363242497"
   ```

**Comment by seanbreckenridge** [comment]
was able to replace a couple of my scripts which previously required python with a bunch of pipes:

https://github.com/seanbreckenridge/HPI/commit/418585ebc2d50323877bbb92e2f71a45ad9b14dd

**Comment by seanbreckenridge** [comment]
The only things that are missing in my mind:

- ability to specify some sort of key on the namedtuple to use as a unique key, and to only print if you haven't already seen it while still in python. that would heavily reduce the amount of items printed by `hpi query`, and remove the need to do a blocking `sort`/`uniq`/`awk '!seen[$0]++'` to remove duplicates while trying to `fzf` against something
- `fzf` integration? last time I checked it was pretty annoying to deal with and the only library that handled it was something using [`plumbum`](https://github.com/tomerfiliba/plumbum) - will see if thats possible to do trivially/implement a prototype

**Comment by karlicoss** [comment]
> ability to specify some sort of key on the namedtuple/data to use as a unique key

yep, that would be cool to figure out... although for most data sources HPI should eliminate duplicates? What's an example where you're running into this issue?

some fzf integration would be super interesting! Guess don't have an opinion whether to keep it separate or not, fine to experiment for now

**Comment by seanbreckenridge** [comment]
Ah, I more meant in the context of doing something like

```
hpi query -s my.browsing.history | jq -r .url
```

... just trying to extract the url or some field from the data

You get lots of duplicates being processed by `jq` because there are duplicate urls since you've visited the same page lots of time

being able to specify `hpi query -s --unique 'url' my.browsing.history`

would mean less mangling the data with additional shell tools and remove blocking `sort`/`uniq` calls

karlicoss:
ah right! feels like there's ought to be some cmdline tool that can handle this... and now I see what you `awk` command is for.. annoying that there isn't a one easy to remember command for it
`jq` also has `unique` operator, but seems it's not stream-oriented https://github.com/stedolan/jq/issues/1555

seanbreckenridge:
There are other tools like [this](https://github.com/lostutils/uq) which handle the edge case, but its annoying that one doesnt exist in coreutils/cant rely on it being installed

seanbreckenridge:
am pretty happy with the query interface for now, will use it for some time and see if there remain to be pain points and see if theres anything else that would be worth adding

PR #172, opened 2021-05-13
smscalls: allow multiple backup dirs

seanbreckenridge:
ran into an issue recently configuring smsbackuprestore, couldn't supply multiple directories since its typed as `PathIsh` instead of `Paths` -- this change is backwards compatible, so I don't think it should be an issue

so my config now looks like:

```python
class smscalls:
    export_path: Paths = ("~/data/SMSBackups", "~/data/phone/SMSBackups")
```

Works fine this way as well:

```python
In [1]: from my.core import get_files

In [2]: from my.config import smscalls

In [3]: smscalls.export_path
Out[3]:
(PosixPath('/home/sean/data/SMSBackups'),
 PosixPath('/home/sean/data/phone/SMSBackups'))

In [4]: get_files(smscalls.export_path, glob='calls-*.xml')
Out[4]:
(PosixPath('/home/sean/data/SMSBackups/calls-20210509005627.xml'),
 PosixPath('/home/sean/data/phone/SMSBackups/calls-20210512161412.xml'))
 ```

 (yes, I only have two files that match, recently removed duplicates)

For context:

I don't sync my entire data directory onto my phone for space concern reasons -- just a subset in ~/data/phone, which includes things like gpslogger or smsbackuprestore files.

`~/data/SMSBackups/` includes old backups I don't want to continue to sync across onto my phone, I move those off periodically

### DIFF
--- a/my/config.py
+++ b/my/config.py
@@ -24,6 +24,9 @@ class hypothesis:
 class instapaper:
     export_path: Paths = ''
 
+class smscalls:
+    export_path: Paths = ''
+
 class pocket:
     export_path: Paths = ''
 
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -5,13 +5,13 @@ Exported using https://play.google.com/store/apps/details?id=com.riteshsahu.SMSB
 
 REQUIRES = ['lxml']
 
-from .core import PathIsh, dataclass
+from .core import Paths, dataclass
 from my.config import smscalls as user_config
 
 @dataclass
 class smscalls(user_config):
-    # directory that SMSBackupRestore syncs XML files to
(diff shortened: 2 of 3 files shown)


seanbreckenridge:
Couldn't find a matching stub config in [my.config](https://github.com/karlicoss/HPI/blob/master/my/config.py), so didn't update that -- perhaps one should be added?

karlicoss:
The stub is mainly so mypy command works -- if you add it & add the module would be great, thanks!
https://github.com/karlicoss/HPI/blob/014494059d64e59af1a96665e4355ce840c00e91/tox.ini#L91-L112

karlicoss:
thanks!

Issue #20, opened 2020-04-14
Find a library for accessing compressed files


karlicoss:
Ideally I want something like [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path), but also supporting archives. Ideally it would be transparent, so for the data consumer it doesn't matter if the file is compressed or not.

At the moment I'm using a [helper](https://github.com/karlicoss/HPI/blob/master/my/kython/kompress.py). It's a bit crude, but mostly works. Would be nice to replace it with a proper library though, that's worth extra dependency.

Potential candidates:

- https://github.com/PyFilesystem/pyfilesystem2
- https://github.com/jaraco/zipp

Some relevant discussions:

- [Make pathlib extensible](https://discuss.python.org/t/make-pathlib-extensible/3428/3)

karlicoss:
This looks interesting https://github.com/PyFilesystem/pyfilesystem2

karlicoss:
Interesting enough, there are some useful functions in the standard library 

https://docs.python.org/3/library/shutil.html#shutil.unpack_archive
https://docs.python.org/3/library/tarfile.html#tarfile.open

However seems that it only supports `.tar` packed files, so not possible to use it as a handy universal file extractor

karlicoss:
Actually `tarfile` isn't too bad, seems that it's capable of guessing compression and the interface is okay. So might be possible to have a short wrapper similar to `ZipPath` if there are more tar.gz exports https://github.com/karlicoss/HPI/commit/5799c062a5e901c2aea1ee9241d7efcfb3dac0d0

karlicoss:
Btw seems that there's been some interesting progress here, some great work towards making pathlib.Path more extensibe (although some thing are only available since python 3.12) https://discuss.python.org/t/make-pathlib-extensible

karlicoss:
After a few years, seems that nothing like this exists, so started extracting in a library https://github.com/karlicoss/kompress

PR #173, opened 2021-05-16
core: add tmp_dir for global access to a tmp dir

seanbreckenridge:
This just adds the tmp_dir attribute so other
HPI modules which want access to some
temporary directory can access this safely
and the user is able to configure this

also added an expanduser() to cache_dir, incase
the user specified something like "~/.cache/",
considering most other uses of PathIsh allow
tildes

discussed this a while back on zulip:
https://memex.zulipchat.com/#narrow/stream/279601-hpi/topic/my.2Ediscord.20.2F.20processing.20archived.20data/near/235445377

### DIFF
--- a/my/core/core_config.py
+++ b/my/core/core_config.py
@@ -82,8 +82,7 @@ class Config(user_config):
         else:
             tpath = Path(tdir)
         tpath = tpath.expanduser()
-        if not tpath.exists():
-            tpath.mkdir(parents=True)
+        tpath.mkdir(parents=True, exist_ok=True)
         return tpath
 
     def _is_module_active(self, module: str) -> Optional[bool]:
--- a/my/core/core_config.py
+++ b/my/core/core_config.py
@@ -42,6 +42,14 @@ class Config(user_config):
     NOTE: you shouldn't use this attribute in HPI modules directly, use Config.get_cache_dir()/cachew.cache_dir() instead
     '''
 
+    tmp_dir: Optional[PathIsh] = None
+    '''
+    Path to a temporary directory.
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
Tangentially related

https://github.com/karlicoss/HPI/issues/20

karlicoss (commented):
Thanks! 
How do we expect it to be used? Something like
```
with TemporaryDirectory(dir=get_tmp_dir()): 
    pass
```
? I guess later we could add a helper thing returning context manager straightaway? so we could write something like
```
with hpi_tmp_dir('discord') as tdir: 
    ....
````

Another thing -- maybe worth using `atexit` hook to clean it up? In case it's used without the context manager.

**Comment by seanbreckenridge** [comment]
I more expected it to be used like the first, like:

```
with TemporaryDirectory(dir=get_tmp_dir()): 
    pass
```

A context manager would make sense, was planning to do this first with my discord module so will see how its typically used there and then PR a helper if the way its used is the same

seanbreckenridge:
Probably no issue with adding an atexit flag?

only concern I have is `shutil.rmtree`'ing something unexpected if something got heavily misconfigured and they werent using the get_tmp_dir accessor

if we add a hpi_tmp_dir contextmanager, maybe could expose that in `core.__init__.py` or `core.common` to encourage use of that instead of importing `core.core_config`

karlicoss:
Yeah agree about removing stuff on accident.. I guess if it'll be used via the context manager, it won't be necessary? Or maybe could start with an atexit warning without deleting, just to be safe

PR #176, opened 2021-07-08
my.core.serialize: simplejson support, more types

seanbreckenridge:
I added a couple extra checks to the default function,
serializing datetime, dates and dataclasses (incase
orjson isn't installed)

(copied from below)

if orjson couldn't be imported, try simplejson
This is included for compatibility reasons because orjson
is rust-based and compiling on rarer architectures may not work
out of the box

as an example, I've been having issues getting it to install
on my phone (termux/android)

unlike the builtin JSON modue which serializes NamedTuples as lists
(even if you provide a default function), simplejson correctly
serializes namedtuples to dictionaries

this just gives another option to people, simplejson is pure python
so no one should have issues with that. (and often they already have
it installed since its used in older python projects for compatability)
orjson is still way faster, so still preferable if its easy and
theres a precompiled build for your architecture (which there typically is)

If you're ever running this with simplejson installed and not orjson,
its pretty easy to tell as the JSON styling is different; orjson has
no spaces between tokens, simplejson puts spaces between tokens. e.g.

```
simplejson: {"a": 5, "b": 10}
orjson: {"a":5,"b":10}
```

### DIFF
--- a/my/core/serialize.py
+++ b/my/core/serialize.py
@@ -1,4 +1,5 @@
 import datetime
+import dataclasses
 from pathlib import Path
 from typing import Any, Optional, Callable, NamedTuple
 from functools import lru_cache
@@ -26,9 +27,13 @@ def _default_encode(obj: Any) -> Any:
         return obj._asdict()
     if isinstance(obj, datetime.timedelta):
         return obj.total_seconds()
+    if isinstance(obj, datetime.datetime) or isinstance(obj, datetime.date):
+        return str(obj)
     # convert paths to their string representation
     if isinstance(obj, Path):
         return str(obj)
+    if dataclasses.is_dataclass(obj):
+        return dataclasses.asdict(obj)
     if isinstance(obj, Exception):
         return error_to_json(obj)
     # note: _serialize would only be called for items which aren't already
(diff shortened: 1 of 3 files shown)


seanbreckenridge:
As dataclasses and datetime is what caused most serialization errors previously, this also just means that `hpi query` should work out of the box for most use cases (though NamedTuples are sorta broken as they're serialized to lists instead; the warning is sent telling you to install orjson/simepljson), even if the user didn't install the optional dependencies

karlicoss:
that's awesome, thanks! Wondering, what are you trying to achieve on the phone, just potentially running same tools as on the computer?

seanbreckenridge:
Yeah, pretty much

I have a few scripts [here](https://github.com/seanbreckenridge/dotfiles/tree/master/.config/i3blocks/blocks) that use `hpi query` to check how much water/food I've eaten in the last day

Other stuff like [my albums](https://github.com/seanbreckenridge/HPI/blob/master/my/albums.py), have a few [aliases](https://github.com/seanbreckenridge/dotfiles/blob/99ad080dad4cbc928c869400c06679bcd02f6bd6/.config/zsh/aliases/other_aliases#L39-L43) to let me know what music to listen to next, which use `hpi query`

seanbreckenridge:
As an update, `orjson` now successfully builds on termux (sometime after https://github.com/ijl/orjson/issues/360)

PR #183, opened 2021-10-31
reddit: swap inheritance order for Protocol

seanbreckenridge:
```
~/Repos/HPI-fork/my/reddit/common.py in <module>
     36
     37 # Note: doesn't include GDPR Save's since they don't have the same metadata
---> 38 class Save(Protocol, RedditBase):
     39     @property
     40     def subreddit(self) -> str: ...

TypeError: Cannot create a consistent method resolution
order (MRO) for bases object, RedditBase
```

Sadly the solution from #179 seems to make mypy happy but doesn't work for the python MRO itself

I think we should revert it for now so that its at least functional and doesn't error, hopefully can find a more concise way to do this.


karlicoss:
ha, so on my [toy example](https://github.com/karlicoss/HPI/pull/179#issuecomment-955779721) I tried swapping `Protocol` and `AP`, and it passes runtime as well now 🤔

```
class BP(AP, Protocol):
    b: int
```

will try it on reddit just in case 😅

seanbreckenridge:
Was trying to figure out if it could always just be put behind a `typing.TYPE_CHECKING` but would probably require another file like `models.py` so that we could still have `_merge_comments` in common -- seems to be another option

PR #186, opened 2021-11-18
docs: add hpi query example, links to other repos

seanbreckenridge:
just some documentation updates

- added an example for `hpi query` to the readme since this was lacking any mention of it
- added the current workaround for converting single file modules to namespaces to module design
- added a list HPI modules repositories and some context on extending a bit


seanbreckenridge:
Was also thinking of making a [cookiecutter](
https://cookiecutter.readthedocs.io/en/1.7.2/) template for an HPI repo, is that something you'd be interested in linking to in the docs?

might streamline people wanting to create their own modules without having to deal the guts of setuptools

Would essentially create a `my` structure with `setuptools.find_namespace_packages` with a README with a bit of instruction on what to do next

seanbreckenridge:
(Failing because of the `my.pdf` changes)

karlicoss:
re: cookiecutter -- yeah, perhaps it makes sense! Not sure how many people are interested, but if it's not too much work that would be great

re: my.pdf -- oh I see -- will take a look

PR #190, opened 2021-12-10
cli/query: allow querying dynamic functions

seanbreckenridge:
As an example of what this could be used for, I keep track of my food here,
https://github.com/seanbreckenridge/ttally/blob/master/ttally/funcs.py

While developing modules for promnesia I realized that
hpi query could be used in any module where you have a top
level function with no arguments which produces some output, and my [HPI body file](https://github.com/seanbreckenridge/HPI/blob/ce9a2f325f096a274fb074d164f7932556716c1b/my/body.py) is just annoying boilerplate that otherwise doesnt need to exist

This does a `getattr` on the module object with the function name, in case there's something in
`globals().setdefault(...)` that resolves the attribute name, or something
like a module-level `__getattr__` which resolves the function name

Doesn't break any existing functionality as this is an edge case where query
might benefit, the default of inspecting the module and filtering for functions
is still done before this `getattr` check

### DIFF
--- a/my/core/query.py
+++ b/my/core/query.py
@@ -54,15 +54,11 @@ def locate_function(module_name: str, function_name: str) -> Callable[[], Iterab
         for (fname, func) in inspect.getmembers(mod, inspect.isfunction):
             if fname == function_name:
                 return func
-        try:
-            # incase the function is defined dynamically,
-            # like with a globals().setdefault(...) or a module-level __getattr__ function
-            func = getattr(mod, function_name)
-            if callable(func):
-                return func
-        except AttributeError:
-            # shouldn't raise an error here, just fall through
-            pass
+        # incase the function is defined dynamically,
+        # like with a globals().setdefault(...) or a module-level __getattr__ function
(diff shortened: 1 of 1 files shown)


my/core/query.py
```diff
@@ -54,6 +54,15 @@ def locate_function(module_name: str, function_name: str) -> Callable[[], Iterab
         for (fname, func) in inspect.getmembers(mod, inspect.isfunction):
             if fname == function_name:
                 return func
+        try:
+            # incase the function is defined dynamically,
+            # like with a globals().setdefault(...) or a module-level __getattr__ function
+            func = getattr(mod, function_name)
+            if callable(func):
+                return func
+        except AttributeError:
```
karlicoss, line 63: hm, wouldn't it be cleaner to `hasattr`, or pass `None` as default to `getattr` and check it later?

seanbreckenridge: Ah, true. I think the `getattr` with `None` works better here because of the additional check to see if its callable

Issue #154, opened 2021-04-03
Allow additional sources for my.location


seanbreckenridge:
Dont expect any changes on this problem any time soon, just creating this issue to track the problem I'm having

Currently I have to overlay `time.tz.via_location`, since I use a [different data source](https://github.com/seanbreckenridge/HPI/blob/master/my/time/tz/via_location.py#L68-L69) (combined from gpslogger, ips from google, facebook, discord etc)

On this repo, it uses `location.google` to grab that info. Slightly unrelated, but I've also parsed the takeout using lxml instead, so [my structure there is different](https://github.com/seanbreckenridge/HPI/tree/master/my/google)

I would prefer if there was a common entrypoint (like `my.location.all`) that could take multiple entrypoints as input, falling back to empty iterators if they aren't enabled/fail to be imported, as that would localize my overlayed changes to the `my.location` package

You can see the current structure for `my.location` [here](https://github.com/seanbreckenridge/HPI/tree/master/my/location)

```
.
├── all.py
├── gpslogger.py
├── ip.py
└── models.py
```

I created this following the discussion we had regarding [merging pushshift data](https://github.com/karlicoss/HPI/issues/89#issuecomment-701688805)

I've also slightly modified the Location NT, so it can track whether this source was from an accurate (e.g. Google or gpslogger) or estimate (geolocation based on ip)

```python
class Location(NamedTuple):
    lng: float
    lat: float
    dt: datetime
    # approximate accuracy
    # true means exact, false means its based on ip/auxiliary info
    accuracy: bool
```

Am a bit conflicted on how to handle this many data sources...

Would need some modifications, would probably create individual files for:

- `google`
- `apple` (locations from gdpr export)
- [`ip-based`](https://github.com/seanbreckenridge/HPI/blob/master/my/location/ip.py) (need inidividual empty fallbacks for `blizzard`, `facebook`, `discord`)
- `gpslogger`

Some of those could stay on my branch if you're not interested in having them here, I think its more important to have the following here:

- a `common.py` file, including:
 - a `Location` NT which all other location providers would convert their NT/DTs to
 - a `merge_locations` function, with the typical `set`/`emitted` behavior

Then, to enable additional `location` providers, I could either just overlay the `all.py` file, including my additional imports -- which probably wouldn't ever have to be changed

Something like what was described in the [comment here](https://github.com/karlicoss/HPI/issues/89#issuecomment-701688805)

If there are no issues you foresee here, I'm willing to implement this at some point in the future.

[...]

PR #192, opened 2021-12-23
my.reddit: enable ci tests/fix legacy config handling

karlicoss:


karlicoss:
@seanbreckenridge wondering what you think about the second commit here https://github.com/karlicoss/HPI/pull/192/commits/f289c6233cefb01dbeefe56e0a7d8a4bc54e689b

Without it, at the moment `my.reddit.all:comments` is failing against the old style config while trying to inherit `unconfig.pushshift`. Still going to be an issue if `pushshift_comment_export` is installed (and you still have old style config), but I guess we can think about it later :) 

I guess something to keep in mind when we'll have more modules using `import_source`.

seanbreckenridge:
Ah, sure, this makes sense. Config is the complicated bit here, so delaying that and failing if the module isn't installed at all is better

Will keep that in mind as I work on #154 (the only module I override on my modules currently)

PR #193, opened 2022-01-14
core: drop py36 support, update docs for reddit

seanbreckenridge:
This fixes references to the my.reddit module (often used as an example) in the docs/error message/docstrings

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,9 +19,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest] # TODO windows-latest??
-        python-version: [3.6, 3.7, 3.8]
-        # seems like 3.6 isn't available on their osx image anymore
-        exclude: [{platform: macos-latest, python-version: 3.6}]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
 
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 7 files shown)


seanbreckenridge:
Seems to be because [pdfannots dropped 3.6 support](https://github.com/0xabu/pdfannots/commits/master)?

karlicoss:
Whoops, sorry, somehow missed the notifications for this! 
Looks good and yeah that's fair, let's drop 3.6. Thanks!

seanbreckenridge:
should I update the docs and the setup.py to bump that to `python_requires='>=3.7'` as well?

karlicoss:
yeah if you don't mind updating other things definitely appreciated!

btw, noticed the python 3.10 commit -- yeah I've run into the same issue I think the other week https://github.com/karlicoss/pockexport/commit/924901fdcdc13c0d9c14d06f68c59b7fbc410bd2 , so I guess for now waiting for that to get fixed

seanbreckenridge:
Seems that the Executor type signature has been updated multiple times and its pretty incompatible across versions

The callable was made a positional only argument, but that [feature isnt available under 3.7](https://www.python.org/dev/peps/pep-0570/), so it just causes a syntax error

I think the easiest thing is to ignore the override for now, and the `#type: ignore[override]` can be removed once `3.7` is at EOL

seanbreckenridge:
I think thats all. Also, I expect this to cause a ripple effect and break anything that installs HPI on 3.6.... [likely need to update this](https://github.com/karlicoss/promnesia/blob/c2d7da626ded5c525c94708273abc245e925cc03/.github/workflows/main.yml#L20)

PR #194, opened 2022-01-27
cli/query: add short flags, stream affects pprint

seanbreckenridge:
adds some short flags as CLI flags for convenience
the --stream flag previously only affected json, but
I can imagine `-o pprint -s -l 5` to print the first
5 items from some function could be useful as well

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -410,7 +410,7 @@ def _locate_functions_or_prompt(qualified_names: List[str], prompt: bool = True)
 def query_hpi_functions(
     *,
     output: str = 'json',
-    stream_json: bool = False,
+    stream: bool = False,
     qualified_names: List[str],
     order_key: Optional[str],
     order_by_value_type: Optional[Type],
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
this is a very tiny breaking change as I [changed this kwarg name](https://github.com/karlicoss/HPI/pull/194/commits/a0526511f2798d67ab9ba062b8e858f623155aa3#diff-234ac9597a5736d0f9e3dff85e89f16be9fc5da9379e653552f2a6f2c6cb0c9dR413), but I can't imagine thats going to affect that many if any people using the `query_hpi_functions` python function in code

karlicoss:
Could also add `**kwargs` and if `kwargs.get('stream_json', None) is not None` then override `stream` with it? But up to you, yeah probably not many people are using it so far

seanbreckenridge:
yeah, I had the thought to use kwargs, but there being so many options/parameters to the query interface, adding `kwargs` could cause people to misspell kwargs that don't exist leading to difficult to track errors, in an already confusing situation.

I think I value the clarity with their being these exact kw-arguments over possibly breaking something which requires a quick change

PR #197, opened 2022-01-27
initial my.zulip module

karlicoss:


seanbreckenridge:
Tangentially related,

For reference, https://github.com/zulip/zulip/issues/2047 has been in discussion for a while, but https://github.com/zulip/zulip/issues/2047 has become pretty stale so I doubt it'll be merged some time soon

>  Would it be useful to look into that and write a script that fetches this data, before I jump back into this PR?

Maybe possible to just use the zulip API in some way to export your own messages, so this may become a module like

```
zulip
├── all.py
├── api_export.py
├── common.py
└── organization.py
```

since not everyone on a zulip has access to an organization export

karlicoss:
Ha, nice, was literally just about to post the same link :) 
Yeah that's the plan! I guess I could rename to `organization`, `org_export` is a bit cryptic? 
And yeah, would be nice to figure out API access, but figured starting with organization takeout is a bit easier.

seanbreckenridge:
Yeah, I think `organization.py` and something like `zulipexport.py` (what that API script/DAL might be called?) would make sense

karlicoss:
Oh, I guess another consideration is that Zulip instances are separate (unlike e.g. Discord where you get all your instances in a single archive), so it's potentially meaningful to reuse zulip module with different source archives? Immediately comes to mind

- just specifying them explicitly in the config (i.e. instead of `Paths`, `Dict[Instance, Paths]`). Easy to implement/use but slightly awkward though and changes the 'shape' of the modules. E.g. from `messages` function do you return `Iterator[Message]` or `Dict[Instance, Iterator[Message]]`? 
- overlays or something (e.g. `my.zulip_memex`, `my.zulip_lean`). Although a little awkward to setup, plus then if you wanna use them in downstream tools (e.g. when I add it to `promnesia`), would also have to write boilerplate for these custom modules

seanbreckenridge:
I think its nicer from a users point of view (though causes a lot of duplication) to just attach the instance to every message, it avoids the return type not 

for example, in discord, I [optionally attach a guild](https://github.com/seanbreckenridge/discord_data/blob/7d4921ab8b2c82a0575fae5bbc0b358b8245884d/discord_data/parse.py#L64-L80)

That would mean every message just has a `Optional[Instance]`, then if you wanted to filter to a particular zulip ream/instance you just filter on that field

could think of it just like a subreddit or a discord server being attached to every message

karlicoss:
Yep, agree, guess `subreddit ->submission -> comment` is a good analogy, it's most sensible to represent it as a uniform stream of comments and then grouping back is fairly easy, if necessary. 

And yeah true about duplication -- kind of a classic normalised/denormalised representation tradeoff. I guess it only really matters/annoying though when you do something like `hpi query` (although there aren't really better interfaces for such sort of thing anyway, except sql)
Another one is probably `cachew` -- the cache might bloat if you keep the whole `Channel` object with every message (and also it likely allocates a lot of them when you deserialise them, even though they are identical). Perhaps this is solvable with some intermediate representation, as an example, `Channel` and `PreMessage` object, which keeps `channel_id` instead of `Channel` objects. Then you could cache both, and the top level `messages` function would load the `PreMessage` objects from cache, find the corresponding `Channel`, build a `Message` from it and emit

karlicoss:
Guess the `Channel`/`PreMessage` could even be saved in the same cache (since cachew supports union types). 
Then the downstream functions just need to process `Iterator[Union[Channel, PreMessage]]`, e.g. `channels()` function could simply filter `Channel` objects out, `messages` would build `Message` as I described above etc.

karlicoss:
another example where used different kind of messages for parsing and more abstract representation -- seems like a relatively straightforward pattern
https://github.com/karlicoss/HPI/pull/201/files

PR #198, opened 2022-01-28
my.zulip: extract Server/Sender objects, experiment with normalised and denormalised objects

karlicoss:


karlicoss:
@seanbreckenridge kind of a followup for this https://github.com/karlicoss/HPI/pull/197#issuecomment-1023701639
not too bad, a little boilerpate during `_Message -> Message` conversion, but since it's all checked by mypy perhaps acceptable

seanbreckenridge:
guess the only other thing to consider is what the structure is with PMs when thats eventually added

maybe `Server.name` is just `Private Message with 'Name'`, but not sure if there'd be an easier way to filter to your PMs with a particular user

karlicoss:
Seems like PMs more likely to do with the `recipient` field and its type https://github.com/zulip/zulip/blob/0c2e4eec200d986a9a020f3e9a651d27216e0e85/zerver/models.py#L1334-L1341

```
$ jq '.zerver_message | map(.recipient)' messages-000001.json | sort | uniq -c
      1 [
      1 ]
    197   637566,
     68   637572,
    369   637574,
      1   637596
    166   637596,
      1   637597,
      1   637598,
      3   637602,
      6   637603,
      1   637604,
      9   637608,
     13   637609,
     17   638759,
      4   648393,
      7   654333,
     49   656421,
```
and from spot checks looks like the messages in the export all have type 2 (STREAM), I'd expect the PMs have type 1 (PERSONAL)
```
$ rg -A 3 'id...637597'
realm.json
50205:      "id": 637597,
50206-      "type_id": 279611,
50207-      "type": 2
50208-    },
$ rg -A 3 'id...637566'
realm.json
50175:      "id": 637566,
50176-      "type_id": 279598,
50177-      "type": 2
50178-    },
$ rg -A 3 'id...654333'
realm.json
50325:      "id": 654333,
50326-      "type_id": 284107,
50327-      "type": 2
50328-    },

```

PR #209, opened 2022-02-09
twitter/archive: update deprecated imports

seanbreckenridge:
while setting up twitter/archive, this sends
the deprecation warnings that my.core.__init__.py
isn't supported anymore -- just updating those
to use .common

Edit; Oh, also set the logger to debug

### DIFF
--- a/my/twitter/archive.py
+++ b/my/twitter/archive.py
@@ -18,7 +18,8 @@ except ImportError as e:
 
 
 from dataclasses import dataclass
-from ..core import Paths, Res, datetime_aware
+from ..core.common import Paths, datetime_aware
+from ..core.error import Res
 
 @dataclass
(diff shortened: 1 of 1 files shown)


karlicoss:
Hmm, `my.core.__init__` should be fine? I guess you meant the `..common` one?

seanbreckenridge:
Ah, yeah -- [the `common.py` here](https://github.com/karlicoss/HPI/pull/209/files#diff-970e02fe8aa90d125c032a93f98fc3582d22d796552f58e0c4e0b61d7e74d0caL39) caused a warning

seanbreckenridge:
mistakenly and made it `debug` by default which means it still printed logs, updated it to warning

#210 related, since people may want to enable those logs

PR #214, opened 2022-02-11
docs: add link to template

seanbreckenridge:
created a cookiecutter template to create a basic structure which supports installing into `my`

https://github.com/seanbreckenridge/HPI-template

Has two basic example modules with/without the typical module boilerplate

https://github.com/seanbreckenridge/HPI-template/tree/master/HPI-%7B%7B%20cookiecutter.username%20%7D%7D/my


karlicoss:
oh nice, thanks! will keep in mind to update if we come up with some more specific guidelines/patters!

Issue #118, opened 2020-12-24
module suggestion: firefox history/bookmarks/etc.


xdrie:
I think a module for accessing Firefox data would be very useful.
This [documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Places/Database) on the Mozilla website details how the data is stored.

I am thinking that a separate script could be used to walk through that database and generate a JSON dump (similarly to how rexport works), and then an HPI module could provide access to that data.

I'm new to this project so I am not yet really familiar with how modules work, but when I have time I will attempt it and submit a PR.

xdrie:
Oh, it looks like one of the contributors to this repo @seanbreckenridge has [already done](https://github.com/seanbreckenridge/ffexport) this!

I wonder if there is already a corresponding HPI module or does it still need to be written?

seanbreckenridge:
I personally dont use bookmarks in the browser (just have a textfile with a [script](https://sean.fish/d/urlhandler) to open stuff), so I haven't written anything to parse that yet.

Otherwise yeah, [ffexport](https://github.com/seanbreckenridge/ffexport) lets you export history, have a script [here](https://github.com/seanbreckenridge/HPI/blob/master/jobs/backup_firefox.job) that saves my history sqlite file every couple weeks.

The [my.browsing](https://github.com/seanbreckenridge/HPI/blob/master/my/browsing.py) file here uses parts of ffexport to load the data in; it also [copies the live history database](https://github.com/seanbreckenridge/HPI/blob/master/my/browsing.py#L55-L62) when computing my so it includes any backups and the current history.

seanbreckenridge:
Just as an update, I've since converted that into [browserexport](https://github.com/seanbreckenridge/browserexport), which supports reading history from:

- Firefox (and Waterfox)
- Chrome (and Chromium, Brave, Vivaldi)
- Safari
- Palemoon

If you wanted to use this, you could install my HPI modules alongside this repository ([see here](https://github.com/seanbreckenridge/HPI#install))

Run `hpi module install my.browsing` to install dependencies

setup a config block in your config file like:

```python
# uses browserexport https://github.com/seanbreckenridge/browserexport
class browsing:
    export_path: Paths = "~/data/browsing"

    # additionally, read use my active firefox database
    from browserexport.browsers.firefox import Firefox

    live_databases: Paths = Firefox.locate_database()
```

Then its just:

```
[ ~ ] $ ipython

In [1]: from my.browsing import history

[...]

PR #216, opened 2022-02-13
browser: parse browser history using browserexport

seanbreckenridge:
from my modules:
https://github.com/seanbreckenridge/HPI/blob/1fba8ccf2fe93d67198aaa77355cac5f5910e7ac/my/browser/export.py

moved the nested configuration blocks above the code generation since it doesnt work properly

### DIFF
--- a/my/browser/common.py
+++ b/my/browser/common.py
@@ -1,0 +1,11 @@
+import os
+from my.core.util import __NOT_HPI_MODULE__
+
+
+def _patch_browserexport_logs():
+    # patch browserexport logs if HPI_LOGS is present
+    if "HPI_LOGS" in os.environ:
+        from browserexport.log import setup as setup_browserexport_logger
+        from my.core.logging import mklevel
+
+        setup_browserexport_logger(mklevel(os.environ["HPI_LOGS"]))
--- a/my/browser/export.py
+++ b/my/browser/export.py
@@ -1,5 +1,5 @@
 """
-Parses Browser history using [[http://github.com/seanbreckenridge/browserexport][browserexport]]
+Parses browser history using [[http://github.com/seanbreckenridge/browserexport][browserexport]]
 """
 
 REQUIRES = ["browserexport"]
--- a/my/browser/export.py
+++ b/my/browser/export.py
@@ -13,74 +13,34 @@ class config(user_config.export):
     # path[s]/glob to your backed up browser history sqlite files
     export_path: Paths
 
-    # paths to sqlite database files which you
-    # use actively, which should be combined into your history
-    # For example:
-    # from browserexport.browsers.all import Firefox
-    # active_databases = Firefox.locate_database()
-    active_databases: Paths
 
-
-import os
 from pathlib import Path
-from typing import Iterator, List
-
-from sqlite_backup import sqlite_backup
+from typing import Iterator, Sequence, List
 
 from my.core import Stats, get_files, LazyLogger
 from my.core.common import mcachew
 
+from browserexport.merge import read_and_merge, Visit
 
-# patch browserexport logs if HPI_LOGS is present
-if "HPI_LOGS" in os.environ:
-    from browserexport.log import setup as setup_browserexport_logger
-    from my.core.logging import mklevel
-
-    setup_browserexport_logger(mklevel(os.environ["HPI_LOGS"]))
+from .common import _patch_browserexport_logs
 
 
 logger = LazyLogger(__name__, level="warning")
 
-
-from browserexport.merge import read_and_merge, merge_visits, Visit
-from browserexport.parse import read_visits
+_patch_browserexport_logs()
 
 
 # all of my backed up databases
-def inputs() -> List[Path]:
-    return list(get_files(config.export_path))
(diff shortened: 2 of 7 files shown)


seanbreckenridge:
related #118

karlicoss (changes requested):

karlicoss:
> not as efficient as [this](https://github.com/karlicoss/promnesia/blob/4f5f7e72b117bf82d0d3b8c388b9cad080b6525e/src/promnesia/sources/browser.py#L42-L79), just does a hash on the backed up databases

ah yeah -- that solution was pretty much messed up anyway, need to think how to support that sort of stuff with cachew..

my/browser/export.py
```diff
@@ -0,0 +1,90 @@
...
+def _active_visits() -> List[Visit]:
+    visits: List[Visit] = []
+    active_dbs = get_files(config.active_databases or "")
+    logger.debug(f"Reading from active databases: {active_dbs}")
+    for ad in active_dbs:
+        conn = sqlite_backup(ad)
+        assert conn is not None
+        try:
+            # read visits, so can close the in-memory connection
+            visits.extend(list(read_visits(conn)))
```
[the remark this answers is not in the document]
seanbreckenridge: yeah, previously this was here because it was using a different method which required to consume the `conn` early else the connection was already closed, but `sqlite_backup` copies the entire db in memory, so works fine without it

my/browser/export.py
```diff
@@ -0,0 +1,90 @@
...
+    return visits
+
+
+Results = Iterator[Visit]
+
+
+# don't put this behind cachew, since the active history database(s)
+# are merged when this is called, whose contents may constantly change
+def history() -> Results:
+    yield from merge_visits([_history_from_backups(), _active_visits()])
```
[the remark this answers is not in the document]
seanbreckenridge: yeah, makes sense

my/browser/common.py
```diff
@@ -0,0 +1,11 @@
...
+from my.core.util import __NOT_HPI_MODULE__
+
+
+def _patch_browserexport_logs():
+    # patch browserexport logs if HPI_LOGS is present
+    if "HPI_LOGS" in os.environ:
+        from browserexport.log import setup as setup_browserexport_logger
+        from my.core.logging import mklevel
+
+        setup_browserexport_logger(mklevel(os.environ["HPI_LOGS"]))
```
seanbreckenridge, line 11: think its fine to have this function in `common.py`? since the imports are in the function anyways

karlicoss (approved):

karlicoss:
thanks!

my/browser/export.py
```diff
@@ -0,0 +1,50 @@
...
+_patch_browserexport_logs()
+
+
+# all of my backed up databases
+def inputs() -> Sequence[Path]:
+    return get_files(config.export_path)
+
+
+def _cachew_depends_on() -> List[str]:
+    return [str(f) for f in inputs()]
```
karlicoss, line 39: @seanbreckenridge do you remember why are we doing it here instead of just passing `depends_on=inputs` to cachew? just tried it and seems to work :thinking: maybe cachew didn't support it back then?

seanbreckenridge: yeah I dont remember exactly why but I think it had to do with testability/name clashes based on a different name for cachew.

I dont think I fully understood it at the time (may still not), I had this helper type `InputSource` so that mypy was happy:

https://github.com/seanbreckenridge/HPI/blob/master/my/utils/input_source.py

So that I could set the default for the `from_paths` in the function signature itself, but still allow the user to override by passing a function which returned a list of paths as inputs:

https://github.com/seanbreckenridge/HPI/blob/aa69d73cc71017b64287a04d917456da8fe5105a/my/bash.py#L64-L70

Something like:

```
from more_itertools import ilen
from my.bash import history, inputs

print(ilen(history(from_paths=lambda: tuple(inputs()[0:1]))))
```

would sort of obviously invalidate the cache, but yeah...

```
 python3 test.py
[INFO    2023-11-06 16:31:38,732 my.bash __init__.py:784 ] [my.bash:history] wrote   59 objects to   cachew (sqlite:/home/sean/.cache/cachew/my.bash:history)
59
```

Not really sure how important that is, someone could always override the inputs by patching their config (having an example of this in the docs might be nice?), I havent really ever found myself specifying a custom function to the inputs so its probably not a feature worthwhile keeping around

you can change it to whatever makes sense for you

seanbreckenridge: I cant reproduce the error anymore but I remember having to rename stuff here specifically:

https://github.com/seanbreckenridge/HPI/blob/aa69d73cc71017b64287a04d917456da8fe5105a/my/bash.py#L64-L70

from `from_paths` to `for_paths` in the cachew helper function. otherwise there was some hard to debug error I couldnt deal with. doesnt seem to happen anymore

karlicoss: yeah good point about documenting patching the config! I was kinda reluctant because it's kinda hacky, but seems like it's working as intended so far
https://github.com/karlicoss/promnesia/blob/master/src/promnesia/sources/browser.py#L53-L67

PR #219, opened 2022-02-21
core/cli: autocomplete module names

seanbreckenridge:
* autocomplete module names for click completion (can hit TAB when typing a module and it lists choices)
* sync new bash completion file (was a bug/typo `click<8.4`)
* bump required click version to support shell completion (was a major rewrite in click 8.0, older versions may not work)


karlicoss:
nice, thanks!

PR #220, opened 2022-03-06
core/source: fix error message, force kwargs for decorator

seanbreckenridge:
typically its only getting `module_name` passed, and the
`default` is left to be `()`, so this should minimize possible
overwriting `default` with the `module_name` while developing

also fixed typo in the error text, and pass both `args` and `kwargs` to the wrapped function


karlicoss:
thanks!

PR #231, opened 2022-04-15
integrating seanbreckenridge/google_takeout_parser

karlicoss:


seanbreckenridge:
Feel free to change the error policy to yield, I had it be drop to simplify code elsewhere (e.g. in location), but should yield them or at least warn when they happen

Also yeah, you can update the `DISABLE_TAKEOUT_CACHE` to be something else, I dont think theres an equivalent in the config, I guess it would be nicer to add things to the takeout config like:

```
@dataclass
class google(user_config):
    disable_takeout_cache: bool
```

Only issue I see with `my/google_takeout.py` is it isnt in a namespace module, so doesnt let users override. Since there are no dependents right now, would be a good time to rename?

`my/google/takeout` is a namespace module, so I think as I suggested [here](https://github.com/karlicoss/HPI/issues/154#issuecomment-1034173704) could rename this file to `my/google/takeout/parser.py`, as that mirrors the name `google_takeout_parser`. As we phase out the html.py/paths.py, it'll just be the single file in the namespace module there, which seems good to me.

karlicoss:
whoa
```
 mypy-misc run-test: commands[1] | hpi module install my.browser.export
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 449, in __init__
    code = compile(arg, '<string>', 'eval')
  File "<string>", line 1
    yield
        ^
SyntaxError: unexpected EOF while parsing

During handling of the above exception, another exception occurred:

  File "/home/runner/work/HPI/HPI/.tox/mypy-misc/bin/hpi", line 33, in <module>
    sys.exit(load_entry_point('HPI', 'console_scripts', 'hpi')())
  File "/home/runner/work/HPI/HPI/.tox/mypy-misc/bin/hpi", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/home/runner/work/HPI/HPI/.tox/mypy-misc/lib/python3.7/site-packages/importlib_metadata/__init__.py", line 203, in load
    module = import_module(match.group('module'))
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/HPI/HPI/my/core/__init__.py", line 12, in <module>
    from .error import Res, unwrap
  File "/home/runner/work/HPI/HPI/my/core/error.py", line 19, in <module>
    ErrorPolicy = Literal["yield", "raise", "drop"]
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 251, in inner
    return cached(*args, **kwds)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 351, in __getitem__
    parameters = tuple(_type_check(p, msg) for p in parameters)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 351, in <genexpr>
    parameters = tuple(_type_check(p, msg) for p in parameters)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 132, in _type_check
    return ForwardRef(arg)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py", line 451, in __init__
    raise SyntaxError(f"Forward reference must be an expression -- got {arg!r}")
SyntaxError: Forward reference must be an expression -- got 'yield'
```
I wonder if it chokes over 'yield' in the Literal somehow...

seanbreckenridge:
Huh weird. Was about to say I have this block as compatibility using typing_extensions so may be the literal but I think I copied that from here

https://github.com/seanbreckenridge/google_takeout_parser/blob/master/google_takeout_parser/compat.py

karlicoss:
yeah there is something similar in HPI, just with an extra hack for runtime
https://github.com/karlicoss/HPI/blob/382f2054298d92077d796f6eccebf467c6f04143/my/core/compat.py#L52-L63

karlicoss:
aaah, so I think it happens not during `mypy`, but during `hpi module install` command (hence it trying to evaluate the code)
then because I declared it as `Union` in runtime, it tries treating its arguments as forward references, and then it crashes

`python3 -c 'from typing import Union; Union[int, "yield"]'` -- this causes the same crash basically, doesn't matter which python version

karlicoss:
okay, I tricked it :) 
I think I'll merge it now, we can think about dealing with cache paths and on/off switches for it later? Might make sense to wait till we have some similar module with its own caching so don't overgenerify now?

seanbreckenridge:
yep, sounds good enough to me

will see how difficult it is to use zippath in my [discord module](https://github.com/seanbreckenridge/HPI/blob/7350e9a8791ee4272c36b261e62f30490d8bb0da/my/discord/data_export.py) and see if theres any duplication we can factor out

karlicoss:
Thanks @seanbreckenridge!
I also compared promnesia modules (with the one from your fork) for the new and old takeouts, after a couple of tweaks the new module is a superset of the old data :) I'll submit a PR to promnesia later

seanbreckenridge:
Great :tada:

Think next thing for me to do will be to make a PR to let `google.takeout.parser` be used as a source for my location (as mentioned [here](https://github.com/karlicoss/HPI/issues/154)), as well as probably merging some version of the [models/all](https://github.com/seanbreckenridge/HPI/tree/7350e9a8791ee4272c36b261e62f30490d8bb0da/my/location) file from my repo, and then finally update the location provider in `time.tz`

At that point [I'll have removed all overrides](https://github.com/seanbreckenridge/HPI/issues/19) (currently override time.tz to use google_takeout_parser) that exist in my repository, my repo will just be additional modules.

Will probably maintain an `all.py` which contains a bunch of personal location sources in my [HPI-personal](https://github.com/seanbreckenridge/HPI-personal) repo, to prevent overriding the default `all.py` file here.

PR #237, opened 2022-04-26
location: add all.py, using takeout/gpslogger/ip

### LINKED ISSUE #154: Allow additional sources for my.location

seanbreckenridge:
Should resolve #154

Zulip convo: https://memex.zulipchat.com/#narrow/stream/279601-hpi/topic/locations

Still need to update the docs, but code has been updated -- can take a look at this whenever you're free

CI may fail, made quite a few updates

Still TODO:

- [x] filter locations in my.location.via_location._locations based on `my.config.time.tz.via_location.require_accuracy`

### DIFF
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -121,7 +121,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
    google takeout (using =my.google.takeout.parser=), with a fallback on
    manually defined home locations.
 
-   You might also be able to use [[file:../my/location/via_ip.py][=my.location.via_ip=]] which uses =my.ip.all= to
+   You might also be able to use [[file:../my/location/via_ip.py][my.location.via_ip]] which uses =my.ip.all= to
    provide geolocation data for an IPs (though no IPs are provided from any
  of the sources here). For an example of usage, see [[https://github.com/seanbreckenridge/HPI/tree/master/my/ip][here]]
 
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -173,7 +173,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
    #+begin_src python
    class time:
        class tz:
-           policy = "convert"
+           policy = 'keep'
 
            class via_location:
                # less precise, but faster
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -99,7 +99,6 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
     Parses browser history using [[http://github.com/seanbreckenridge/browserexport][browserexport]]
 
     #+begin_src python
-    @dataclass
     class browser:
         class export:
             # path[s]/glob to your backed up browser history sqlite files
@@ -126,7 +125,6 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
  of the sources here). For an example of usage, see [[https://github.com/seanbreckenridge/HPI/tree/master/my/ip][here]]
 
     #+begin_src python
-    @dataclass
     class location:
         home = (
              # supports ISO strings
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -21,7 +21,7 @@ If you have some issues with the setup, see [[file:SETUP.org::#troubleshooting][
   - [[#myreddit][my.reddit]]
   - [[#mybrowser][my.browser]]
   - [[#mylocation][my.location]]
-  - [[#mytimetzvialocation][my.time.tz.via.location]]
+  - [[#mytimetzvia_location][my.time.tz.via_location]]
   - [[#mypocket][my.pocket]]
   - [[#mytwittertwint][my.twitter.twint]]
   - [[#mytwitterarchive][my.twitter.archive]]
@@ -93,6 +93,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
             export_path: Paths
 
     #+end_src
+   
 ** [[file:../my/browser/][my.browser]]
 
     Parses browser history using [[http://github.com/seanbreckenridge/browserexport][browserexport]]
@@ -111,8 +112,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
             # active_databases = Firefox.locate_database()
             export_path: Paths
     #+end_src
-
- ** [[file:../my/location][my.location]]
+** [[file:../my/location][my.location]]
 
    Merged location history from lots of sources.
 
@@ -148,8 +148,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
               # guess ~15km accuracy for IP addresses
               accuracy: float = 15_000
     #+end_src
-
- ** [[file:../my/time/tz/via_location.py][my.time.tz.via_location]]
+** [[file:../my/time/tz/via_location.py][my.time.tz.via_location]]
 
    Uses the =my.location= module to determine the timezone for a location.
 
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -160,7 +160,7 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
     - you might trust your original timezone, or it might just be UTC, and you want to use something more reasonable
 
    #+begin_src python
-   Policy = Literal[
+   TzPolicy = Literal[
        'keep'   , # if datetime is tz aware, just preserve it
        'convert', # if datetime is tz aware, convert to provider's tz
        'throw'  , # if datetime is tz aware, throw exception
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -16,9 +16,12 @@ If you have some issues with the setup, see [[file:SETUP.org::#troubleshooting][
 - [[#toc][TOC]]
 - [[#intro][Intro]]
 - [[#configs][Configs]]
-  - [[#mygoogletakeoutpaths][my.google.takeout.paths]]
+  - [[#mygoogletakeoutparser][my.google.takeout.parser]]
   - [[#myhypothesis][my.hypothesis]]
   - [[#myreddit][my.reddit]]
+  - [[#mybrowser][my.browser]]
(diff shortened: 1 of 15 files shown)


my/time/tz/via_location.py
```diff
@@ -71,12 +98,13 @@ def _iter_local_dates(start=0, stop=None) -> Iterator[DayWithZone]:
...
-def most_common(l):
-    res, count = Counter(l).most_common(1)[0] # type: ignore[var-annotated]
+def most_common(lst: List[DayWithZone]) -> DayWithZone:
+    res, _ = Counter(lst).most_common(1)[0]  # type: ignore[var-annotated]
     return res
 
 
-@mcachew(cache_path=cache_dir())
+# refresh _iter_tzs once per day -- don't think a better depends_on is possible dynamically
+@mcachew(logger=logger, depends_on=lambda: str(date.today()))
```
seanbreckenridge, line 107: Not sure about this, but otherwise it just never updates as far as I understand

seanbreckenridge:
I think thats everything I could think of, let me know if you have any changes

seanbreckenridge:
Also, not using `geopy` anywhere, that was used in `my.location.google` but didnt want to copy that to everywhere, perhaps a separate `my.location.geo` could import from `my.location.all` to enrich the data, if you want elevation/additional geolocation info?

otherwise feels like its polluting all the other sources

seanbreckenridge:
Ah, I guess [this section](https://github.com/karlicoss/HPI/pull/237/files#diff-5bdb66c06c27fd59b29c9202a64666fbd5281e080c72d97b0b3aed060a203e8fR93-R97) assumed that the dates were sorted, but thats not really possible since `my.location.all` could have any order of dates, from multiple sources.

It works when you just have the one google source, but since this combines a bunch of (possibly unknown and unsorted), its dropping a bunch of data there....

Guess the check for 'local time goes backwards' should just be removed -- the way Im 'solving' this is to sort the local dates, in `_iter_tzs`, so at least we can group by sorted locations (as thats required)

my/time/tz/via_location.py
```diff
@@ -46,20 +53,43 @@ class DayWithZone(NamedTuple):
...
+        zone = finder.timezone_at(lat=lat, lng=lon)
         if zone is None:
-            warnings.append(f"Couldn't figure out tz for {l}")
+            warnings.append(f"Couldn't figure out tz for {lat}, {lon}")
             continue
         tz = pytz.timezone(zone)
         # TODO this is probably a bit expensive... test & benchmark
-        ldt = l.dt.astimezone(tz)
+        ldt = dt.astimezone(tz)
         ndate = ldt.date()
```
seanbreckenridge, line 93: I think this section may need to be removed? Since we cant control the order of my.location

So, this will be slower, but thats the trade off for allowing any sort of location data as input

Its not trivial to cache the `my.location.all` function either, since it depends on any amount of inputs, so creating an accuracte cachew depends_on would be difficult

my/location/gpslogger.py
```diff
@@ -0,0 +1,74 @@
...
+
+def inputs() -> Sequence[Path]:
+    return get_files(config.export_path, glob="*.gpx")
+
+
+def _cachew_depends_on() -> List[float]:
+    return [p.stat().st_mtime for p in inputs()]
+
+
+# TODO: could use a better cachew key/this has to recompute every file whenever the newest one changes
```
karlicoss, line 44: heh, synthetic export to the rescue :)

my/location/common.py
```diff
@@ -0,0 +1,17 @@
...
+from my.core import __NOT_HPI_MODULE__
+
+DateIsh = Union[datetime, date, str]
+
+LatLon = Tuple[float, float]
+
+
+# TODO: add timezone to this? can use timezonefinder in tz provider instead though
+class Location(NamedTuple):
+    lon: float
```
karlicoss, line 13: Kinda nit, but it's inconsistent with `LatLon` above?
Not sure if should change, was never sure which one is right, ran into this recently https://news.ycombinator.com/item?id=30228981 and still confused haha. So just flagging in case you have an opinion

my/time/tz/via_location.py
karlicoss: ouch. yeah interesting, something must have been lost during refactoring -- I might have not used this for a while

my/time/tz/via_location.py
karlicoss: Hmm yeah, I mean in principle we could either demand the upstream providers to provide it as sorted (and for many it'll already be, like google takeout?).
Or alternatively sort before merging them in `all` (via a helper `merge` function), then the whole thing would be much more iterative. But I guess can think about it later

karlicoss (commented):
Looks good, thanks! A few discussion points

my/time/tz/via_location.py
karlicoss: Or could even do some magic, e.g. function which takes multiple generator factories (functions `() -> Iterator`), assumes that they are emitting sorted values, and tries merging with https://docs.python.org/3/library/heapq.html#heapq.merge 

If something out of order is detected, it starts from scratch with the offending factory wrapped in `sorted` (hence merging factories, not plain `Iterator`s). I guess realistically it would very quickly lead to detecting which iterables are already sorted and which aren't, usually the order is either increasing, decreasing or just completely random? 

Just leaving it here, but potentially might be useful for other data sources too, e.g. often it's kinda nice to get things according to timestamps

doc/MODULES.org
```diff
@@ -108,6 +112,81 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
...
+   datetimes in UTC, to prevent confusion whether or not its a local
+   timezone, one from UTC, or one in your timezone.
+
+   Depending on the specific data provider and your level of paranoia you might expect different behaviour.. E.g.:
+    - if your objects already have tz info, you might not need to call localize() at all
+    - it's safer when either all of your objects are tz aware or all are tz unware, not a mixture
+    - you might trust your original timezone, or it might just be UTC, and you want to use something more reasonable
+
+   #+begin_src python
+   Policy = Literal[
```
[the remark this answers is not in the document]
seanbreckenridge: Just in the docs? Or should I rename the underlying Literal as well?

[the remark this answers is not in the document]
karlicoss: Yeah, maybe worth renaming in code as well (with a backwards compatible `Policy = TzPolicy`)?

my/time/tz/via_location.py
seanbreckenridge: Yeah, use a heapmerge would make sense in situations like these, but I still think it might be prone to errors because a user might just overwrite `all.py` with something else (like I've done [here](https://github.com/seanbreckenridge/HPI-personal/blob/master/my/location/all.py)) and then the magic doesn't work anymore. As usual, the magic might make it nice for us, but its not super extendible -- I feel like changing an `all.py` file shouldnt break a separate module

For now I think I'll go with the iterative approach of sorting the `_locations` in `_iter_tzs` -- since its cached when its used repeatedly

seanbreckenridge: Another idea could be in `all.py` modules to specify the sources as a `SOURCES = [ ]` key at the toplevel, so then we could actually do some sort of magic and sort the iterators by querying that, instead of doing it in `all.py` itself. Otherwise, `my.location.all.locations()` is just a function and no way to get what the underlying iterators are. If the `SOURCES` var doesn't exist, then fallback to sorting the giant list of locations iteratively

And then locations in `my.location.all` would just be:

```
for provider in SOURCES:
    yield from provider()
```

seanbreckenridge:
believe thats all the changes -- `_iter_tz` now works for me properly (previously was dropping a bunch of locations since it was out of order)

karlicoss:
nice! thanks

PR #239, opened 2022-05-01
doctor: better quick option propogation for stats

seanbreckenridge:
* use contextmanager for quick stats instead of editing global state directly
* send quick to lots of stat related functions, so they could possibly be used without doctor, if someone wanted to
* if a stats function has a 'quick' kwarg, send the value there as well
* add an option to sort locations in my.time.tz.via_location

### DIFF
--- a/my/core/stats.py
+++ b/my/core/stats.py
@@ -13,12 +13,12 @@ from .common import StatsFun, Stats, stat
 
 # TODO maybe could be enough to annotate OUTPUTS or something like that?
 # then stats could just use them as hints?
-def guess_stats(module_name: str) -> Optional[StatsFun]:
+def guess_stats(module_name: str, quick: bool=False) -> Optional[StatsFun]:
     providers = guess_data_providers(module_name)
     if len(providers) == 0:
         return None
     def auto_stats() -> Stats:
-        return {k: stat(v) for k, v in providers.items()}
+        return {k: stat(v, quick=quick) for k, v in providers.items()}
     return auto_stats
 
 
--- a/doc/MODULES.org
+++ b/doc/MODULES.org
@@ -189,6 +189,10 @@ For an extensive/complex example, you can check out ~@seanbreckenridge~'s [[http
                # less precise, but faster
                fast: bool = True
 
+               # sort locations by date
+               # incase multiple sources provide them out of order
+               sort_locations: bool = True
+
                # if the accuracy for the location is more than 5km (this
                # isn't an accurate location, so shouldn't use it to determine
                # timezone), don't use
--- a/my/config.py
+++ b/my/config.py
@@ -84,6 +84,7 @@ class time:
     class tz:
         class via_location:
             fast: bool
+            sort_locations: bool
             require_accuracy: float
 
 
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -215,11 +215,11 @@ def modules_check(*, verbose: bool, list_all: bool, quick: bool, for_modules: Li
         verbose = True
     vw = '' if verbose else '; pass --verbose to print more information'
 
-    from . import common
-    common.QUICK_STATS = quick # dirty, but hopefully OK for cli
-
     tabulate_warnings()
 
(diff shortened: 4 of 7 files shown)


my/core/stats.py
```diff
@@ -13,12 +13,12 @@
 
 # TODO maybe could be enough to annotate OUTPUTS or something like that?
 # then stats could just use them as hints?
-def guess_stats(module_name: str) -> Optional[StatsFun]:
+def guess_stats(module_name: str, quick: bool=False) -> Optional[StatsFun]:
```
karlicoss, line 16: hmm, I feel like `quick` argument belongs to the `auto_stats` below instead? since it's a property of the stats function, not the module itself
then it would be inferred by `inspect.signature` later

karlicoss (changes requested):

my/core/stats.py
seanbreckenridge: Oh, I guess it could be -- but that also means the user has to do what happens in `doctor` with the `inspect.signature` (which isn't obvious if they're don't look at the code there)  I was just giving the user more options in case they were calling guess_stats manually -- they could force quick to be true by doing:

```
guess_stats(mod, quick=True)
```

or they could:

```
with quick_stats():
    guess_stats(mod)
```

since in that case in `_stat_iterable` would then look at the global `QUICK_STATS`

karlicoss: ah, makese sense. I thought of `guess_stats` is kind of for internal usage tbh? but don't mind either way, just curious what use do you see outside of the doctor

seanbreckenridge: yeah would agree its mostly for internal use, but I could see someone wanting to make a custom script which uses only particular modules to test the functionality of

e.g. if I wanted to use doctor on a particular subset of modules programmatically by picking which `my` namespace package

```

In [26]: for mod in list(_discover_path_importables(list(sorted(my.__path__))[3], "my")):
    ...:     mm = importlib.import_module(mod.name)
    ...:     stats_func = guess_stats(mod.name, quick=True)
    ...:     print(stats_func())
    ...: 
{'transactions': {'transactions': {'count': '100+'}}}
{'history': {'history': {'count': '100+'}}, 'to_listen': {'to_listen': {'count': '100+'}}}
{'achievements': {'achievements': {'count': 26, 'last': datetime.datetime(2021, 1, 18, 20, 53)}}, 'forum_posts': {'forum_posts': {'count': '100+', 'last': datetime.datetime(2014, 7, 17, 18, 2, 57, tzinfo=datetime.timezone.utc)}}}
{'locations': {'locations': {'count': '100+', 'last': datetime.datetime(2021, 4, 26, 4, 1, 42, 766000, tzinfo=datetime.timezone.utc)}}}
```

Could always do a bash script and list stuff by then converting to something like:

```
$ fd --glob '*.py' my | tr '/' '.' | rcut -d'.' -f2- | xargs hpi doctor -Sq
✅ OK  : my.location.all                                   
✅     - stats: {'locations': {'count': '100+', 'last': datetime.datetime(2021, 4, 26, 4, 1, 42, 766000, tzinfo=datetime.timezone.utc)}}
✅ OK  : my.mint                                           
✅     - stats: {'_all_balances': {'count': '100+', 'last': datetime.datetime(2020, 10, 20, 12, 5, 52, tzinfo=<git.objects.util.tzoffset object at 0x7f64c7edc340>)}, 'transactions': {'count': '100+'}}
✅ OK  : my.nextalbums                                     
✅     - stats: {'history': {'count': '100+'}, 'to_listen': {'count': '100+'}}
✅ OK  : my.old_forums                                     
✅     - stats: {'forum_posts': {'count': '100+', 'last': datetime.datetime(2014, 7, 17, 18, 2, 57, tzinfo=datetime.timezone.utc)}, 'achievements': {'count': 26, 'last': datetime.datetime(2021, 1, 18, 20, 53)}}
```

But thought might as well make it a bit more user friendly if someone wanted to use it manually for a couple modules from python instead of from the cli

PR #252, opened 2022-06-04
zulip: add stats function

seanbreckenridge:
mostly so that this is included in `hpi doctor` output, since it having the stats func means its detected as a module

### DIFF
--- a/my/zulip/organization.py
+++ b/my/zulip/organization.py
@@ -170,3 +170,11 @@ def messages() -> Iterator[Res[Message]]:
             )
             continue
         assert_never(x)
+
+
+from my.core import Stats
+def stats() -> Stats:
+    from my.core import stat
+    return {
+        **stat(messages)
+    }


seanbreckenridge:
```
$ hpi doctor -S my.zulip.organization
✅ OK  : my.zulip.organization                             
✅     - stats: {'messages': {'count': 912, 'last': datetime.datetime(2022, 1, 25, 8, 20, 5, 225768)}}
```

karlicoss:
ah makes sense! I think I'm too used to running doctor directly on specific modules (so module discovery isn't triggered because of this https://github.com/karlicoss/HPI/blob/master/my/core/__main__.py#L250), and stat worked before it's inferred here https://github.com/karlicoss/HPI/blob/b5f266c2bd2cb463f97bc68c627d754d6e7b6377/my/core/__main__.py#L278

I guess eventually might be nice for module discovery to benefit from the same stat inference logic? That way won't have to add extra stats functions, unless we're trying to do something special?

seanbreckenridge:
Yeah, I often run a `hpi doctor` periodically on everything just to make sure nothings broken with new data that may have been added, and saw it wasnt being picked up.

Only issue I see with trying to move the doctor logic to `util.py` is that that is all using the `discovery_pure` and scanning `my.__path__` to find module names, but if its possible to traverse the AST and find the stats functions or something, could be nice to do in the future

PR #255, opened 2022-06-05
allow installing modules in bulk, add experimental --parallel flag

karlicoss:
`time tox -e mypy-misc` (removed the actual mypy call)

before (each module in a separate 'hpi install' command)
```
real	1m45.901s
user	1m19.555s
sys	0m5.491s
```

in a single 'hpi install' command (multiple modules)
```
real	1m31.252s
user	1m6.028s
sys	0m5.065s
```

single 'hpi install' command with --parallel
```
real	0m15.674s
user	0m50.986s
sys	0m3.249s
```

Almost 8x speedup!

Sadly doesn't help that much with github actions times, perhaps because the runner only has one core

before
```
2022-06-05T21:32:06.8525970Z mypy-misc create: /home/runner/work/HPI/HPI/.tox/mypy-misc
2022-06-05T21:34:51.3006446Z Success: no issues found in 35 source files
```

after
```
2022-06-05T23:06:25.8043510Z mypy-misc create: /home/runner/work/HPI/HPI/.tox/mypy-misc
2022-06-05T23:08:35.6433602Z Success: no issues found in 35 source files
```

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -1,6 +1,7 @@
 import functools
 import importlib
 import inspect
+from itertools import chain
 import os
 import shutil
 import sys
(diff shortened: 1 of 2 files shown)


karlicoss:
Also as suggested [here](https://pythonspeed.com/articles/faster-pip-installs/#going-faster-sometimes-with-disabled-compilation), tried `--no-compile` option for `pip` (mypy doesn't need `.pyc` files anyway), but it doesn't really give any significant speedup. Perhaps most of the time is spent cloning git repos..

PR #259, opened 2022-10-15
core/serialize: serialize stdlib Decimal class

seanbreckenridge:
updates the `dumps` fallback to serialize the [builtin decimal](https://docs.python.org/3/library/decimal.html) class

chose to not convert it to a float since that might
lose precision, and likely the whole reason youre
storing it as a Decimal is to avoid that

### DIFF
--- a/my/core/serialize.py
+++ b/my/core/serialize.py
@@ -1,6 +1,7 @@
 import datetime
 import dataclasses
 from pathlib import Path
+from decimal import Decimal
 from typing import Any, Optional, Callable, NamedTuple
 from functools import lru_cache
 
@@ -36,6 +37,11 @@ def _default_encode(obj: Any) -> Any:
         return dataclasses.asdict(obj)
     if isinstance(obj, Exception):
         return error_to_json(obj)
+    # if something was stored as 'decimal', you likely
+    # don't want to convert it to float since you're
+    # storing as decimal to not lose the precision
+    if isinstance(obj, Decimal):
+        return str(obj)
     # note: _serialize would only be called for items which aren't already
     # serialized as a dataclass or namedtuple
     # discussion: https://github.com/karlicoss/HPI/issues/138#issuecomment-801704929


seanbreckenridge:
CI failing because of unrelated mypy issue:

```
$ mypy -p my.core
my/core/common.py: note: In function "_identity":
my/core/common.py:72: error: A function returning TypeVar
should receive at least one argument containing the same
TypeVar  [type-var]
    def _identity(v: T) -> V:
                           ^
Found 1 error in 1 file (checked 30 source files)
```

seanbreckenridge:
I think this fixes it since then `T` and `V` is in context:

```diff
diff --git a/my/core/common.py b/my/core/common.py
index a4dd4c9..3a2df47 100644
--- a/my/core/common.py
+++ b/my/core/common.py
@@ -69,18 +69,19 @@ def group_by_key(l: Iterable[T], key: Callable[[T], K]) -> Dict[K, List[T]]:
     return res
 
 
-def _identity(v: T) -> V:
-    return cast(V, v)
-
 
 # ugh. nothing in more_itertools?
 def ensure_unique(
         it: Iterable[T],
         *,
         key: Callable[[T], K],
-        value: Callable[[T], V]=_identity,
+        value: Optional[Callable[[T], V]]=None,
         key2value: Optional[Dict[K, V]]=None
 ) -> Iterable[T]:
+    if value is None:
+        def _identity(v: T) -> V:
+            return cast(V, v)
+        value = _identity
     if key2value is None:
         key2value = {}
     for i in it:
@@ -113,7 +114,7 @@ def make_dict(
         it: Iterable[T],
         *,
         key: Callable[[T], K],
-        value: Callable[[T], V]=_identity
+        value: Optional[Callable[[T], V]]=None,
 ) -> Dict[K, V]:
     res: Dict[K, V] = {}
     uniques = ensure_unique(it, key=key, value=value, key2value=res)
```

Though Im not totally clear on what the code is doing, so unsure if this breaks anything

karlicoss:
Thanks, looks good!
re: mypy failure -- I figured it's just easier to ignore it on `_identity` considering it's an internal function https://github.com/karlicoss/HPI/commit/7098d6831f37667f7dd874704f23bf594abfa198

PR #261, opened 2022-11-12
google.takeout.parser: recreate cache on upgrade

seanbreckenridge:
Related PR in `google_takeout_parser`: https://github.com/seanbreckenridge/google_takeout_parser/pull/37

Whenever the `google_takeout_parser` pip version changes, the hash changes, so it recomputes

Went with the basic solution of prepending the google takeout pip version to the `depends_on` (though could be changed to whatever in the future if cachew exposes a version kwarg to store in that table instead)

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -376,7 +376,9 @@ def module_install(*, user: bool, module: Sequence[str], parallel: bool=False) -
     ]
 
     cmds = []
-    if parallel:
+    # disable parallel on windows, sometimes throws a
+    # '[WinError 32] The process cannot access the file because it is being used by another process'
+    if parallel and sys.platform not in ['win32', 'cygwin']:
         # todo not really sure if it's safe to install in parallel like this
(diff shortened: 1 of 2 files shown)


seanbreckenridge:
was some #`[WinError 32] The process cannot access the file because it is being used by another process` on the windows CI, so I thought it might be the parallel `hpi module install` on windows, which seems to have been the case

seanbreckenridge:
@karlicoss may have missed the notification for this? not sure

karlicoss:
ah sorry, yep, missed the original one somehow! 
Thanks looks good -- I've been wondering how to make it more generic (e.g. maybe cachew could take in package name and extract the version by itself?). But good start for now :)

seanbreckenridge:
yeah that sounds good could also do something like pass it a string or a callable. If its a string look for a `__version__` attribute or default to `unknown`, if its a callable then it runs the function to return the version

PR #275, opened 2023-02-28
add example for denylist, update ci

seanbreckenridge:
pushed denylist [to my branch](https://github.com/seanbreckenridge/HPI-personal/blob/master/my/ip/all.py) now that this is merged here, added that to the docs

just a small fix to link to that and fix module names for `hpi module install` in the CI (didnt do that in the last PR to avoid possible merge conflicts)


seanbreckenridge:
Yeah wow, is totally unpredictable, I had previously disabled CI on windows because of a `Win32Error`, perhaps the issue this is unpredictable is because theres some race condition of it checking for the install when installing in parallel?

pasted the error in `__main__.py` as well:

`# WARNING: Error parsing requirements for sqlalchemy: [Errno 2] No such file or directory: '/Users/runner/work/HPI/HPI/.tox/mypy-misc/lib/python3.7/site-packages/SQLAlchemy-2.0.4.dist-info/METADATA'`

Issue #266, opened 2023-02-04
add semantic location history to my.location.google_takeout


seanbreckenridge:
just added semantic location history to `google_takeout_parser`: https://github.com/seanbreckenridge/google_takeout_parser/pull/38

https://github.com/seanbreckenridge/google_takeout_parser/blob/9cf5386f74b0b337e24dd8368691e8f6b36aaa83/google_takeout_parser/models.py?rgh-link-date=2023-02-04T04%3A57%3A32Z#L143

Can probably also convert them down to `Location`, but will make a decision after the location fallback restructure #263 

will PR this once thats done, just adding this issue here so I don't forget

seanbreckenridge:
Added in #278

PR #278, opened 2023-03-04
google_takeout: add semantic location history

seanbreckenridge:
google_takeout_parser already de-normalizes the `timelineObject` onto `PlaceVisit` (duplicating `placeVisit` data from each `timelineObject` onto the `PlaceVisit` dataclass), so we can just 1-1 map from those objects to here

https://github.com/seanbreckenridge/google_takeout_parser/blob/9cf5386f74b0b337e24dd8368691e8f6b36aaa83/google_takeout_parser/parse_json.py#L166-L168

No docs from google, but did find [this community-run docs page](https://locationhistoryformat.com/reference/semantic/#/properties/timelineObjects/items/) for the location history format

```
$ hpi doctor -S my.location.google_takeout_semantic
✅ OK  : my.location.google_takeout_semantic               
✅     - stats: {'locations': {'count': 1676, 'last': datetime.datetime(2022, 11, 2, 2, 58, 31, tzinfo=datetime.timezone.utc)}}
```

### DIFF
--- a/my/config.py
+++ b/my/config.py
@@ -89,6 +89,13 @@ class location:
         export_path: Paths = ''
         accuracy: float
 
+    class google_takeout_semantic:
+        # a value between 0 and 100, 100 being the most confident
+        # set to 0 to include all locations
+        require_confidence: float = 40
+        # default accuracy for semantic locations
+        accuracy: float = 100
+
 
 from my.core.compat import Literal
 class time:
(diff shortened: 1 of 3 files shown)


seanbreckenridge:
Resolves #266

karlicoss:
nice, thank you!

PR #279, opened 2023-03-05
cli completion doc updates, hide legacy import warning

seanbreckenridge:
- core/cli: hide warnings when autocompleting
- link to completion in setup/troubleshooting
- update completion docs to make source path clear


seanbreckenridge:
Previously if I tried to tab complete, it prints the warning every time:

```
[ ~ ] $ hpi module install my./home/sean/Repos/HPI-karlicoss/my/core/legacy.py:50: UserWarning: importing my.reddit is DEPRECATED! Instead, import from my.reddit.rexport or my.reddit.all See https://github.com/karlicoss/HPI/blob/master/doc/MODULE_DESIGN.org#allpy for more info.

  W.high(f'''\
        hpi module install my.t
 ```

karlicoss:
thanks!

PR #286, opened 2023-04-10
cli: add option to output locations to gpx files

seanbreckenridge:
This adds an `--output gpx` option to `hpi query` that will output anything that
looks like a `LocationProtocol` to GPX on stdout

Found a nice GUI tool to quickly preview GPX files: https://github.com/tumic0/GPXSee

```bash
hpi query my.location.gpslogger --recent 1w -o gpx >out.gpx; gpxsee out.gpx
```

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -542,6 +542,14 @@ def query_hpi_functions(
                 pprint(item)
         else:
             pprint(list(res))
+    elif output == 'gpx':
+        from my.location.common import locations_to_gpx
+
+        # can ignore the mypy warning here, locations_to_gpx yields any errors
+        # if you didnt pass it something that matches the LocationProtocol
+        for exc in list(locations_to_gpx(res, sys.stdout)):   # type: ignore[arg-type]
+            click.echo(str(exc), err=True)
+        sys.stdout.flush()
     else:
         res = list(res)  # type: ignore[assignment]
         # output == 'repl'
@@ -681,7 +689,7 @@ def module_install_cmd(user: bool, parallel: bool, modules: Sequence[str]) -> No
 @click.option('-o',
               '--output',
               default='json',
-              type=click.Choice(['json', 'pprint', 'repl']),
+              type=click.Choice(['json', 'pprint', 'repl', 'gpx']),
               help='what to do with the result [default: json]')
 @click.option('-s',
               '--stream',
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -547,7 +547,7 @@ def query_hpi_functions(
 
         # can ignore the mypy warning here, locations_to_gpx yields any errors
         # if you didnt pass it something that matches the LocationProtocol
-        for exc in list(locations_to_gpx(res, sys.stdout)):   # type: ignore[arg-type]
+        for exc in locations_to_gpx(res, sys.stdout):   # type: ignore[arg-type]
             click.echo(str(exc), err=True)
         sys.stdout.flush()
     else:
--- a/my/location/common.py
+++ b/my/location/common.py
@@ -1,5 +1,5 @@
 from datetime import date, datetime
-from typing import Union, Tuple, Optional
+from typing import Union, Tuple, Optional, Iterable, TextIO, Iterator
 from dataclasses import dataclass
 
 from my.core import __NOT_HPI_MODULE__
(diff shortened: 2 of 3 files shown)


seanbreckenridge:
Ah, looks like `lxml` type info was updated, so this now fails since `.get` is optional?:

Should I change the `NamedTuple` definitions to reflect that?

```
my/smscalls.py:45: error: Argument 1 to "_parse_dt_ms" has incompatible type
"Optional[str]"; expected "str"  [arg-type]
                dt=_parse_dt_ms(cxml.get('date')),
                                ^~~~~~~~~~~~~~~~
my/smscalls.py:46: error: Argument "dt_readable" to "Call" has incompatible
type "Optional[str]"; expected "str"  [arg-type]
                dt_readable=cxml.get('readable_date'),
                            ^~~~~~~~~~~~~~~~~~~~~~~~~
my/smscalls.py:47: error: Argument 1 to "int" has incompatible type
"Optional[str]"; expected
"Union[str, bytes, bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer, SupportsInt, SupportsIndex, SupportsTrunc]"
 [arg-type]
                duration_s=int(cxml.get('duration')),
                               ^~~~~~~~~~~~~~~~~~~~
                               ```

**Comment by seanbreckenridge** [comment]
rebased on top of the mypy fix from smscalls to see if CI works for this (since adding the `gpxpy` dependency optionally)

will fix later today

**Comment by seanbreckenridge** [comment]
Oh, and also on the

> # hmm -- would it be useful to allow the user to split this into tracks?, perhaps by date?

I found something like https://github.com/tkrajina/gpx-cmd-tools/blob/master/gpxtools/gpxsplitter.py if thats something we want to pursue, perhaps with localized datetimes, or by clustering locations together for a particular trip etc.

File: my/location/common.py [hunk 1]
```diff
@@ -32,3 +32,48 @@ class Location(LocationProtocol):
...
+        try:
+            point = gpxpy.gpx.GPXTrackPoint(
+                latitude=location.lat,
+                longitude=location.lon,
+                elevation=location.elevation,
+                time=location.dt,
+                comment=location.datasource,
+            )
+        except AttributeError:
+            yield TypeError(
```
karlicoss, line 73: there is also `@runtime_checkable` but never tried it myself, and it needs `typing_extensions`. So just in case you haven't seen it :)  https://mypy.readthedocs.io/en/stable/protocols.html#using-isinstance-with-protocols

my/core/__main__.py
```diff
@@ -542,6 +542,14 @@ def query_hpi_functions(
                 pprint(item)
         else:
             pprint(list(res))
+    elif output == 'gpx':
+        from my.location.common import locations_to_gpx
+
+        # can ignore the mypy warning here, locations_to_gpx yields any errors
+        # if you didnt pass it something that matches the LocationProtocol
+        for exc in locations_to_gpx(res, sys.stdout):   # type: ignore[arg-type]
```
karlicoss, line 550: I guess one downside is that it doesn't respect the same exception policies as select_range, and just dumps everything to terminal. But not a big deal, and perhaps can think about it later if necessary

my/location/common.py
seanbreckenridge: Ah yeah, I saw that but disregarded because of the version requirement. I guess this would work even if the user duck-typed some other location class :+1:

my/core/__main__.py
seanbreckenridge: ah yeah, this was more just a safeguard to prevent crashing, can add the two if statements to handle errors

karlicoss: one (somewhat hacky?) way to make it fit into the existing mechanism might be hacking a wrapper iterable before calling `select_range`, e.g.
```
if output == 'gpx':
    input_src = map(check_conforms_to_location_procotol, input_src)

# select_range call etc
```

karlicoss: Ah I see what you mean -- that way exceptions will end up in stdout and mess with gpx output. Anyway I think it's fine as it is :) But yeah later might worth adding an option to `select_range`/cli support to log exceptions to stderr?

seanbreckenridge: yeah sounds good, will do that when I PR some docs for `hpi query`

PR #287, opened 2023-04-11
smscalls: make some fields optional

seanbreckenridge:
reflects the new types-lxml package
https://github.com/abelcheung/types-lxml

### DIFF
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -18,45 +18,54 @@ config = make_config(smscalls)
 
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import NamedTuple, Iterator, Set, Tuple
+from typing import NamedTuple, Iterator, Set, Tuple, Optional
 
 from lxml import etree # type: ignore
 
-from .core.common import get_files, Stats
+from my.core.common import get_files, Stats
+from my.core.error import Res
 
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
If you think this is fine, will make a PR to promnesia afterwards to fix the source there

karlicoss:
thanks!

karlicoss:
uhh I should get more sleep :roll_eyes:
Thinking again about it -- is this genuinely the case for smscalls module that these fields can be none? 
If not perhaps best to keep the model stricter and just add asserts to make mypy happy? So the downstream users don't have to handle `None` where it's not really necessary.
Haven't used smscalls for a while so not sure myself.

seanbreckenridge:
yeah, I still use this module

unless the android app changes their format I guess there should never be errors?

so if you prefer I can revert this and instead yield an error if the keys dont exist, that way its stricter

I dont see the sms backup app changing their format after all these years anyways

karlicoss:
Not a huge deal but imo no need to make it optional without a reason. Yielding should be fine, although I guess would require a change in promnesia due to mypy. But perhaps best to always assume modules can yield exception for consistency, so it'll be a useful change anyway :)

PR #289, opened 2023-04-15
smscalls: make checking for keys stricter

seanbreckenridge:
sort of reverts #287, but also makes some other improvements

this allows us to remove some of the Optional's to
make downstream consumers easier to write. However,
this keeps the return type as a Res (result, with errors),
so downstream consumers will have to handle those incase
the schema ever changes (highly unlikely)

also added the 'call_type/message_type' with a comment
there describing the values

I left 'who' Optional I believe it actually should be -
its very possible for there to be no contact name, added
a check incase its '(Unknown)' which is what my phone
sets it to

### DIFF
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -3,6 +3,8 @@ Phone calls and SMS messages
 Exported using https://play.google.com/store/apps/details?id=com.riteshsahu.SMSBackupRestore&hl=en_US
 """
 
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
I left `call_type`/`message_type` an integer, could make it an Enum (but not a huge fan of pythons enum) or a `Literal` if you'd prefer

karlicoss:
awesome, thanks! Yeah same, not really a fan of python enums

PR #290, opened 2023-04-15
query: add --warn-exceptions, dateparser, docs

seanbreckenridge:
added --warn-exceptions (like --raise-exceptions/--drop-exceptions, but
lets you pass a warn_func if you want to customize how the exceptions are
handled. By default this creates a logger in main and logs the exception

added dateparser as a fallback if its installed (it's not a strong dependency, but
I mentioned in the docs that it's useful for parsing dates/times)

added docs for query, and a few examples

--output gpx respects the --{drop,warn,raise}--exceptions flags, have
an example of that in the docs as well

### DIFF
--- a/my/core/query_range.py
+++ b/my/core/query_range.py
@@ -73,13 +73,28 @@ def parse_datetime_float(date_str: str) -> float:
         return ds_float
     try:
         # isoformat - default format when you call str() on datetime
+        # this also parses dates like '2020-01-01'
         return datetime.fromisoformat(ds).timestamp()
     except ValueError:
         pass
     try:
         return isoparse(ds).timestamp()
     except (AssertionError, ValueError):
-        raise QueryException(f"Was not able to parse {ds} into a datetime")
+        pass
+
+    try:
+        import dateparser # type: ignore[import]
+        # dateparser is a bit more lenient than the above, lets you type
+        # all sorts of dates as inputs
+        # https://github.com/scrapinghub/dateparser#how-to-use
+
+        res: Optional[datetime] = dateparser.parse(ds)
+        if res is not None:
+            return res.timestamp()
+    except ImportError:
+        pass
+
+    raise QueryException(f"Was not able to parse {ds} into a datetime")
 
 
 # probably DateLike input? but a user could specify an order_key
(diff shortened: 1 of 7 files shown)


seanbreckenridge:
Oh also, moved some of the error iterators `drop_exeptions`,`raise_exceptions` from query to `my.core.error`, I find myself using those pretty often so I think would be useful to make those 'public'

seanbreckenridge:
Alright, have read over this a few times, should be good to review

my/core/query_range.py
```diff
@@ -73,13 +73,28 @@ def parse_datetime_float(date_str: str) -> float:
...
     except ValueError:
         pass
     try:
         return isoparse(ds).timestamp()
     except (AssertionError, ValueError):
-        raise QueryException(f"Was not able to parse {ds} into a datetime")
+        pass
+
+    try:
+        import dateparser # type: ignore[import]
```
karlicoss, line 86: when I used dateparser in the past, I'd usually configure it via settings to prefer YMD/DMY order -- by default it's locale dependent which may be an issue, e.g.

```
dateparser.parse('04-05-2021')
Out[3]: datetime.datetime(2021, 4, 5, 0, 0)
```
which is odd even though I have `en_GB` locale (so I'd expect the default to be DMY). But dunno what's a good default really

my/core/query_range.py
```diff
@@ -73,13 +73,28 @@ def parse_datetime_float(date_str: str) -> float:
...
+        # dateparser is a bit more lenient than the above, lets you type
+        # all sorts of dates as inputs
+        # https://github.com/scrapinghub/dateparser#how-to-use
+
+        res: Optional[datetime] = dateparser.parse(ds)
+        if res is not None:
+            return res.timestamp()
+    except ImportError:
```
karlicoss, line 94: nit, but probably worth wrapping only import in the `except` statement, then the rest of the code can be in `else` block

my/core/query_range.py
seanbreckenridge: Probably a good idea to make it `YMD` by default since thats what we show in examples, and so its possibly easier to debug if someone is getting incorrect dates

karlicoss:
thanks! great docs too!

PR #302, opened 2023-08-09
core/cli: allow user to bypass PEP 668

seanbreckenridge:
when installing dependencies with 'hpi module install',
this now lets a user pass '--break-system-packages' (or '-B'),
which passes the same option down to pip, to allow the user
to bypass PEP 668 and install packages that could possibly
conflict with system packages.

### DIFF
--- a/my/core/__main__.py
+++ b/my/core/__main__.py
@@ -362,7 +362,7 @@ def module_requires(*, module: Sequence[str]) -> None:
         click.echo(x)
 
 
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
heh... bit surprised this worked after 2 months of not running a CI, I was expecting I'd have to fix some mypy issue that had cropped up due to new stubs somewhere

karlicoss:
thanks!

PR #304, opened 2023-08-26
core/cli: update vendored completion files

seanbreckenridge:
update the vendored completion files (click has since pushed
some updates that change the output of the completions)

update required click version to 8.1
so we dont regenerate the vendored completions
wrong in the future

### DIFF
--- a/misc/completion/fish/hpi.fish
+++ b/misc/completion/fish/hpi.fish
@@ -1,9 +1,5 @@
 function _hpi_completion;
-    set -l response;
-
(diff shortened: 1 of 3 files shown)


seanbreckenridge:
`ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/home/runner/work/HPI/HPI/.tox/mypy-misc/lib/python3.8/site-packages/charset_normalizer-3.2.0.dist-info/top_level.txt'`

Hm, may have been a race condition with `the --parallel` flag? Could retry before investigating further

karlicoss:
Sorry, forgot to actually rerun the CI! Looks good now, seems like it's the --parallel flag indeed

Issue #305, opened 2023-09-06
HPI_LOGS envvar no longer works


seanbreckenridge:
Seems [got removed when hpi switched to colorlog](https://github.com/karlicoss/HPI/commit/661714f1d9d56f8c422fb2a12578ee04ccc96329#diff-390a6ba268d0a76aad1f6a97ff3626b88731f24559ad6ea668c67ce0e956886cL41-L44), is there plans to move it to some part of the `my.config.core`?

Otherwise I like having the ability to quickly introspect the cachew hash using it, can make a PR to re-add

seanbreckenridge:
Ah, I guess theres the `LOGGING_LEVEL_` prefixes now, but not sure if it lets you configure everything. The `--debug` flag in the cli hasnt changed [here](https://github.com/karlicoss/HPI/blob/c283e542e3457ecd778fb09e54e725d67104a49a/my/core/__main__.py#L599-L601)

Not sure if  `LOGGING_LEVEL`  should be added as a environment variable, or maybe `LOGGING_LEVEL_GLOBAL`?

karlicoss:
Ah sorry I've been meaning to document but got distracted! I'll take a look later today!

seanbreckenridge:
Am glad to PR some changes the google takeout parser/browserexport modules which previously used the `HPI_LOGS` to update those packages' log levels as well (I'd probably just use LazyLogger to create the logger for each module and then check the loggers computed level to)

Though, not sure what to do about the `--debug` [flag here](https://github.com/karlicoss/HPI/blob/c283e542e3457ecd778fb09e54e725d67104a49a/my/core/__main__.py#L591), I guess could just do a:

```
import .logging as log_module
log_module.DEFAULT_LEVEL = 'DEBUG'
```

Alternatively, could change that line in the logging module to `DEFAULT_LEVEL = os.environ.get("LOGGING_LEVEL", "INFO")`, that would act pretty similarly to `HPI_LOGS` in the past (could even add a check for `HPI_LOGS` to keep backwards compatibility)

karlicoss:
So there are a few things here (none of which I properly documented, sorry :see_no_evil: )

**HPI changes**
- added a helper to get logging level from an env variable, so it's easy to adjust per-module
  https://github.com/karlicoss/HPI/blob/c283e542e3457ecd778fb09e54e725d67104a49a/my/core/logging.py#L78-L81
- only set logger level if it's not been set before
[...]

PR #307, opened 2023-09-06
my.core.logging: compatibility with HPI_LOGS

seanbreckenridge:
I had tried to set the `DEFAULT_LEVEL` like we discussed in #305 here:

https://github.com/karlicoss/HPI/blob/c283e542e3457ecd778fb09e54e725d67104a49a/my/core/logging.py#L49

But if the module had already set a level like:

```python
from my.core import make_logger

logger = make_logger(__name__, level="warning")
```

then any attempt to use `HPI_LOGS` (which would use the `DEFAULT_LEVEL`) would be ignored, which is not what the user would expect (they'd expect the `--debug` flag to just override everything)

So, added a check after the `LOGGING_LEVEL_` check in the `get_env_level` function instead

### DIFF
--- a/doc/MODULE_DESIGN.org
+++ b/doc/MODULE_DESIGN.org
@@ -233,3 +233,17 @@ The main goals are:
 It could be argued that namespace packages and editable installs are a bit complex for a new user to get the hang of, and this is true. But fortunately ~import_source~ means any user just using HPI only needs to follow the instructions when a warning is printed, or peruse the docs here a bit -- there's no need to clone or create your own override to just use the ~all.py~ file.
 
 There's no requirement to use this for individual modules, it just seems to be the best solution we've arrived at so far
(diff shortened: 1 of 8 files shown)


karlicoss:
Thanks, looks good! 
Only thing I'm thinking is that `LOGGIN_LEVEL*` and `HPI_LOGS` have inconsistent naming. Perhaps we could use `LOGGING_LEVEL_HPI` as the primary control instead, and use `HPI_LOGS` as a legacy fallback? Not a huge deal though.

seanbreckenridge:
Yeah, I think I mentioned the same in #305, can update the docs to something more consistent

seanbreckenridge:
Should be good, added a `my.core.warnings.medium` incase the user had `HPI_LOGS`

PR #314, opened 2023-09-30
location: make accuracy default config floats

seanbreckenridge:
related: https://github.com/karlicoss/cachew/issues/54


seanbreckenridge:
weird, looks like pep517 related error: https://github.com/karlicoss/HPI/actions/runs/6365711590/job/17282936363#step:5:70

karlicoss:
Perhaps it's the parallel install once again? Anyway I retried and it worked.

karlicoss:
Thanks! I wonder if there is a setting in mypy we could use to catch these? Fells like it's more often than not we want to avoid these coercions, although I might underestimate how often they are used

PR #324, opened 2023-10-21
core/stats: less duplication in output, report stats of first item, report input stats

karlicoss:


my/core/stats.py
```diff
@@ -12,16 +12,42 @@
...
+def test_guess_stats() -> None:
+    from datetime import datetime
+    import my.core.tests.auto_stats as M
+
+    auto_stats = guess_stats(M.__name__)
+    assert auto_stats is not None
+    res = auto_stats()
+
+    assert res == {
+        'inputs': {
```
karlicoss, line 38: @seanbreckenridge any objections on enabling stats for inputs/sources? We disabled it here https://github.com/karlicoss/HPI/commit/d71383ddee171035ee5bd6cfdf335031fc48a7a8 , but I don't remember why, I imagine it was noisy to only have input count? I think with first/last items it would be a bit more useful so I added it back

seanbreckenridge: I think its fine to have it in stats if we want it, I had just assumed we didn't. Probably fine to add it yeah, would be nice to keep track of input counts in my [weekly snapshot](https://github.com/seanbreckenridge/HPI-personal/blob/master/jobs/all/doctor_snapshot.job)

PR #356, opened 2023-12-03
gpslogger: catch broken xml file error

seanbreckenridge:
```
File "/home/sean/.local/lib/python3.11/site-packages/gpxpy/parser.py", line 146, in parse
    raise mod_gpx.GPXXMLSyntaxException('Error parsing XML: %s' % str(e), e)
gpxpy.gpx.GPXXMLSyntaxException: Error parsing XML: Extra content at the end of the document, line 39, column 22 (<string>, line 39)
```

not sure exactly what the cause for it is, but sometimes if my laptop loses power (so it crashes) it adds a bunch of weird EOF characters to the end of the file, and this ends up breaking the parser.

I can manually fix that, just thought Id prevent any broken XML files from crashing the parser.

### DIFF
--- a/my/location/gpslogger.py
+++ b/my/location/gpslogger.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from typing import Iterator, Sequence, List
 
 import gpxpy
+from gpxpy.gpx import GPXXMLSyntaxException
 from more_itertools import unique_everseen
 
 from my.core import Stats, LazyLogger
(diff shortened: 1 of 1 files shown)


karlicoss:
Sorry, missed the notification! I guess ideally we'd make it yield Exception or something instead, but for now doesn't hurt to merge as is. Thanks!

PR #370, opened 2024-05-31
smscalls: parse mms from smscalls export

seanbreckenridge:
parses all the MMS file/content parts, left comments alongside me exploring the data object

want to actually try consuming the output of this somewhere before I think it should be merged

also, there are some items parsed as MMS which are actually just text (like, theyre SMIL (XML-format) containers which contain a single `text/plain` file), with no pictures attached. So, its just a text thats stored in a MMS container, in an ideal world we can filter those out and add them to `messages`? But may not always work perfectly

So, we should probably at least note that mms can have text items, or... maybe combine them/have some helper `.property` methods on the `MMS` that give you nicer filtering/output?

will post some examples when I'm playing with this, gonna go sleep now

```bash
$ hpi doctor -S my.smscalls
✅ OK  : my.smscalls                                      
✅     - stats: {..... 'mms': {'count': 2491, 'first': datetime.datetime(2019, 12, 18, 4, 22, 46, tzinfo=datetime.timezone.utc), 'last': datetime.datetime(2024, 5, 29, 1, 31, 19, tzinfo=datetime.timezone.utc)}}
```

### DIFF
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -240,9 +240,11 @@ def _extract_mms(path: Path) -> Iterator[Res[MMS]]:
                 user_type = addr_data.get('type')
                 if user_address is None or user_type is None:
                     addr_str = etree.tostring(addr_parent).decode()
-                    return RuntimeError(f'Missing one or more required attributes [address, type] in {addr_str}')
+                    yield RuntimeError(f'Missing one or more required attributes [address, type] in {addr_str}')
+                    continue
                 if not user_type.isdigit():
-                    return RuntimeError(f'Invalid type {user_type}')
+                    yield RuntimeError(f'Invalid string {user_type} {type(user_type)}, cannot convert to number')
+                    continue
                 addresses.append((user_address, int(user_type)))
 
         content: List[MMSContentPart] = []
--- a/my/smscalls.py
+++ b/my/smscalls.py
@@ -154,37 +154,67 @@ class MMSContentPart(NamedTuple):
     sequence_index: int
     content_type: str
     filename: str
-    text: str
+    text: Optional[str]
+    data: Optional[str]
 
 
-# https://www.synctech.com.au/sms-backup-restore/fields-in-xml-backup-files/
 class MMS(NamedTuple):
     dt: datetime
     dt_readable: str
-    content: List[MMSContentPart]
+    parts: List[MMSContentPart]
+    # NOTE: these is often something like 'Name 1, Name 2', but might be different depending on your client
     who: Optional[str]
+    # NOTE: This can be a single phone number, or multiple, split by '~' or ','. Its better to think
+    # of this as a 'key' or 'conversation ID', phone numbers are also present in 'addresses'
     phone_number: str
+    addresses: List[Tuple[str, int]]
     # 1 = Received, 2 = Sent, 3 = Draft, 4 = Outbox
     message_type: int
+
+    @property
+    def from_user(self) -> str:
+        # since these can be group messages, we can't just check message_type,
+        # we need to iterate through and find who sent it
+        # who is CC/'To' is not obvious in many message clients
+        #
+        # 129 = BCC, 130 = CC, 151 = To, 137 = From
+        for (addr, _type) in self.addresses:
+            if _type == 137:
+                return addr
+        else:
+            # hmm, maybe return instead? but this probably shouldnt happen, means
+            # something is very broken
+            raise RuntimeError(f'No from address matching 137 found in {self.addresses}')
+
+    @property
+    def from_me(self) -> bool:
+        return self.message_type == 2
 
 
 def mms() -> Iterator[Res[MMS]]:
     files = get_files(config.export_path, glob='sms-*.xml')
 
-    emitted: Set[Tuple[datetime, Optional[str]]] = set()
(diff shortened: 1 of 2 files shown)


seanbreckenridge:
Probably instead of adding it to messages, we can just add helpers...? If there are multiple content parts, pretty common case is for there to be one `text/plain` and multiple `image/jpeg` attached.

Maybe a property that checks if theres only one part and its text, with no other parts attached

it might be nice to be able to handle the common mime types as well, like parsing images into [`PIL` images](https://pillow.readthedocs.io/en/stable/reference/Image.html)

heres what I have:

```
$ hpi query my.smscalls.mms -s | jq '.content.[].content_type' -r | tally
      1 text/x-vcard
      4 video/3gpp
      5 image/gif
      7 text/x-vCard
     43 image/png
    429 image/jpeg
   2405 text/plain
```

Other todos:

- [ ] Want to see if I can decode the target for the translated 'Liked an image'/'Liked a message' RCS stuff

seanbreckenridge:
On the `Liked an Image` stuff being transcribed, looks like a probably not

seems that stuff often breaks when things are re-imported or when you move sim cards, so its likely some translation happening in-app and its not saved perfectly in an export

Was looking to see if I could match any ids like in here:

`<mms date="1648436193000" rr="129" sub="null" ct_t="application/vnd.wap.multipart.mixed" read_status="null" seen="1" msg_box="1" address="<REDACTED>" sub_cs="null" resp_st="null" retr_st="128" d_tm="null" text_only="1" exp="null" locked="0" m_id="mavodi-6-89-1e8-8-ba-628c370c-7d583aca2b" st="null" retr_txt_cs="null" retr_txt="null" creator="com.google.android.apps.messaging" date_sent="1648436191" read="1" m_size="243" rpt_a="null" ct_cls="null" pri="null" sub_id="2" tr_id="null" resp_txt="null" ct_l="<REDACTED>" m_cls="null" d_rpt="129" v="18" _id="740" m_type="132" readable_date="Mar 27, 2022 7:56:33 PM" contact_name="<REDACTED>">
    <parts>
    <part seq="0" ct="text/plain" name="null" chset="3" cd="null" fn="null" cid="&lt;0&gt;" cl="null" ctt_s="null" ctt_t="null" text="Loved an image"/>
  </parts>
</mms>`

but my basic exploration of grepping IDs across the file doesnt seem to have worked. Theres lots of random key/values/ids though, maybe it could be indexes in conversation/some auto-indexed ID...? cant figure it out, just guessing

seanbreckenridge:
Ahhh, any group convo is converted into a MMS since otherwise it cant accurately encode who sent which message.

So thats why there are messages that are 'just text'

`message_type` is just `# 1 = Received, 2 = Sent, 3 = Draft, 4 = Outbox`, for group messages `contact_name` is just a list -- so you just know if you recieved it but dont know from who. The `addresses` described here https://www.synctech.com.au/sms-backup-restore/fields-in-xml-backup-files/ 

```
<addr address="<--->" type="130" charset="106"/>
<addr address="<--->" type="130" charset="106"/>
<addr address="<--->" type="130" charset="106"/>
<addr address="<--->" type="130" charset="106"/>
<addr address="<--->" type="137" charset="3"/>
<addr address="<--->" type="151" charset="106"/>
```

where it specifies which number and then a type tells you who actually sent it...

`129 = BCC, 130 = CC, 151 = To, 137 = From` ... weird schema.

seanbreckenridge:
Was able to use it nicely in some scripts I have to preview convos, and save any images found to ~/.cache/sms-images, synced those scripts up here:

https://github.com/seanbreckenridge/HPI-personal/commit/d8b0539cb294e15f3188c1ba66ee0772d63af534

am sure these some small issues I may have missed, but those will get found with more usage - I think this is good enough to merge and start using.

seanbreckenridge:
@karlicoss  this should be good to review/merge

karlicoss:
Thanks! I literally never received or sent an MMS, so don't really have any option, happy to merge :)

seanbreckenridge:
> never recieved or sent an MMS

I didnt think I had that many either, but any group chat turns itself into an MMS because theres no other way to encode who the message is coming from when there are multiple people (see [comment](https://github.com/karlicoss/HPI/pull/370#issuecomment-2141132488))...

 So, apparently I had a couple hundred because am in group chats with family etc.

I mostly care about this for random images that are stored in it though, am going to embed the correct date into that and add it to my eventual photos module.

thanks :+1:

seanbreckenridge:
Oh, you may have to [edit this line then](https://github.com/karlicoss/HPI/pull/370/files#diff-70bbe7f962d2cbb2218de8fc4f5da701e576b86188aa1d14dd0d7a08d67867a9R10), I assumed you had a few so I just put the standard, eh, "check if theres 10 things in this" test

karlicoss:
Yeah, noticed this as well, but sadly this isn't running regularly anyway. Ideally we'd add some test file to https://github.com/karlicoss/hpi-testdata or something like that, so it can run on CI

seanbreckenridge:
Yeah, I always get paranoid about pushing location data test files or a modified xml export just because I always think Ive missed something, I've been wanting to write some tool that takes JSON/XML as input and creates valid-looking dummy-data.

PR #409, opened 2024-11-22
doc: some performance notes for query_range

purarue:


purarue:
was just debugging query_range perf for a while, thought I'd leave some notes for anyone who happens to be doing the same in the future

purarue:
also includes some fixes for `ruff`, since there was a version bump that caused CI to fail

karlicoss:
thanks!

PR #411, opened 2024-12-23
smscalls: allow large XML files as input

purarue:
once XML files increase past a certain size
(was about 220MB for me), the parser just throws
an error because the tree is too large (iirc for
security reasons)

could maybe look at using iterparse in the future
to parse it without loading the whole file, but this
seems to fix it fine for me


karlicoss:
220MB xml is impressive :skull: thanks!

purarue:
yeah, images really increase the size of the export :sweat_smile: 

I think at some point I will do a rewrite and write a test to do a regression test against the current parsing mechanism

I can no longer parse it on my phone, I get a lxml memory overflow error

PR #427, opened 2025-03-08
workaround to make `pip install hpi[testing]` work while dependency group support is missing in pip

karlicoss:
see

- https://github.com/karlicoss/HPI/commit/d4246f4b65951fcf97946eec604f2e6ecf506985#commitcomment-153450319
- https://github.com/pypa/pip/pull/13065

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ optional = [
     "enlighten",  # for CLI progress bars
 ]
 
-[dependency-groups]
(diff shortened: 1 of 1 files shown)


purarue:
Have you found uv to be a good experience?

Would you recommend using it?

karlicoss:
Yep, definitely recommend! Mainly for virtualenv-like stuff, quickly testing packages without having to install them etc. Super easy to test on different python versions, that sort of thing.

PR #460, opened 2025-09-06
doc: mention how to install the testing dependency group

purarue:


purarue:
as an aside, did end up migrating most of items over to pyproject.toml, have pushed an __init__.py to my personal repo for path ordering:

https://github.com/purarue/HPI-personal/blob/master/my/__init__.py

PR #490, opened 2026-07-16
core: fix warning when using relative path in MY_CONFIG

karlicoss:
Supersedes https://github.com/karlicoss/HPI/pull/486

Thanks @yutotakano for the original report/fix!


yutotakano:
Yay!!! Thank you so much!

PR #495, opened 2026-07-19
my.moves: add Moves export provider

AndrewSB:
## What changed

- add `my.moves`, a standard-library-only provider for original Moves export ZIPs
- expose source-faithful `Day`, `Segment`, `Activity`, and `TrackPoint` wrappers with complete raw mappings and archive/member provenance
- provide flattened iterators and content-free aggregate statistics
- add HPI configuration and module-list entries
- add a runtime-generated, entirely synthetic nested-ZIP test fixture covering ordering, alternate-format exclusion, unknown/null fields, optional timestamps, error isolation, provenance, and no-write behavior

## Why

Moves is defunct, but its historical exports remain valuable location and activity archives. Daily storyline JSON is the richest unique record set; selecting only that representation avoids duplicates from the weekly, monthly, yearly, full, and alternate-format views in the same export.

The provider deliberately does not resolve Moves data against Google, infer visits or workouts, or create a derived database. It exposes the original source records so downstream HPI modules can make those decisions separately.

## Impact

Users can configure one immutable outer export ZIP and query days, segments, activities, and track points through normal HPI conventions. A malformed daily member is yielded as an exception without hiding valid members before or after it.

## Validation

- `pytest -q src/my/tests/moves.py src/my/core/tests/test_config.py` — 19 passed, 1 existing skip
- Ruff check and format check
- mypy and ty checks
- content-free smoke test against a complete real export: 1,691 days, 30,499 segments, 42,959 activities, 418,334 track points, 0 parser errors


AndrewSB:
@purarue can you comment on whether you think this should be upstreamed, or if it should instead be something I maintain in a fork? 

My moves export is from November 16, 2013–July 22, 2018, and has 418,334 GPS track points and 30,499 storyline segments across 1,141 dates. I'd like to be able to include this information in `my.location`

AndrewSB:
spoke to @purarue, I'll maintain this in my own fork

AndrewSB:
Thanks for the guidance on module ownership and namespace packages. I’ve moved this provider into a standalone repository based on the HPI template: https://github.com/AndrewSB/HPI-moves

It installs alongside upstream HPI under `my.moves`, retains the synthetic test suite, and its CI is passing. I’m closing this draft in favor of maintaining the module there.

